PR #5, opened 2018-07-23
Add lessons about the permanent Web

vmx:
The lessons start at `/#/11` and go up to `/#/16`.

This [robably only make sense with the DAG being displayed (as it is mostly about links). I push it, so that we can discuss if it makes sense to polish or abandon it.

The lessons work, just details (like the image) is missing. So feel free to follow along.

It's not that much about coding, it's rather about building up a DAG. I had this idea before we had the workshopper tool as a real thing.

### DIFF
--- a/src/components/Home.vue
+++ b/src/components/Home.vue
@@ -11,6 +11,28 @@
         <router-link to="/02">Lesson 2: TBD</router-link>
       </li>
     </ul>
+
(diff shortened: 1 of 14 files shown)


mikeal:
There's a couple adjustments I want to make but I'm going to merge it first and then get it fixed.

PR #14, opened 2018-07-25
WIP: blog lessons

vmx:
Notes:

 - Currently the pre-populated code is the solution of the previous
   lesson. So you can e.g. copy the code from lesson 3 into the code
   box of lesson 2 and it should validate correctly.
 - Validations for lesson 1-5 is done
 - Validations for other lessons is missing
 - I'm still unclear on how to validate lesson 6

@olizilla Here's the PR, just for you!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16100,6 +16100,11 @@
         "safe-buffer": "^5.0.1"
       }
     },
+    "shallow-equal": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/shallow-equal/-/shallow-equal-1.0.0.tgz",
+      "integrity": "sha1-UI0YOLPeWQq4dXsBGyXkMJAJRfc="
+    },
     "shebang-command": {
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/shebang-command/-/shebang-command-1.2.0.tgz",
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "ipfs-css": "^0.6.0",
     "marked": "^0.4.0",
     "raw-loader": "^0.5.1",
+    "shallow-equal": "^1.0.0",
     "tachyons": "^4.11.1",
     "vue": "^2.5.16",
     "vue-monaco-editor": "0.0.19",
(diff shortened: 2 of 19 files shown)


olizilla:
@vmx I did a run through, here are my minimal notes. Hopefully they make sense, but we can talk them over tomorrow if that'd help.

## Q1
need to direct the user to **edit the existing blog post objects**
need to remind that links are an object with just a slash. I forgot them and the error 
```
"fail: 'A blog post is missing the link to the author."
```
didn't help me figure out the solution

## Q3
too many tags... 3 is enough

## Q4
need to be super clear that the expected content is 'dogs'

## Q5
need to fix question text.
need to be super specific about expected order of posts in a tag.

## Q6 
need to fix question text.
need to remind about the shape of links.
listing the posts newest first would make more sense to me.

## Q7 
need to introduce ipfs.dag.get.
example code has a bug
```
ReferenceError: computers is not defined at run
```
`computers` should be `computerPostCid`

Then i hit an error:
```
Error: No resolver found for codec "undefined" at IPLDResolver._get
```
which I think is because ipfs.dag.get is expecting a CID object... i'm trying...
```js
  async function getPrevs (cid, res) {
    res = res || []
    const {value} = await ipfs.dag.get(cid)
    res.push(value)
    if (!value.prev) return res
    return getPrevs(value.prev, res)
  }

  return getPrevs(dogPostCid)
```
...aha!

```js
  async function getPrevs (cid, res) {
    res = res || []
    const {value} = await ipfs.dag.get(cid)
    res.push(value)
    if (!value.prev) return res
    return getPrevs(value.prev['/'], res)
  }

  return getPrevs(dogPostCid.toBaseEncodedString())
```

gonna need to remind people that you have to walk that `'/'`

I'd be super impressed if anyone gets that working. Without being able to see the shapes of the objects getting returned or console.log things it's really hard. We might need to demonstrate that when we explain how `ipfs.dag.get` works. Oh i guess console.log would have appeared in the dev tools... but I didn't think to check. Would be good to add a hint about that too!

vmx:
@olizilla Thanks a lot, that's great feedback. I'll look into it tomorrow. I had the hope that things might make more sense with an explorer seeing the links.

mikeal:
Is this ready to merge now?

olizilla:
I'm just trying out the exercises now

vmx:
@mikeal Coding wise it's ready. I currently work on better lesson descriptions. Once @olizilla is done, he merges. The better descriptions is a subsequent PR.

olizilla:
@vmx a wording issue on the error message for the last challenge... replace

'your function cannot be executed with an error'

'your function threw an error'

Can be done as part of a subsequent pr tho.

PR #28, opened 2018-07-30
feat: expand the exercise

olizilla:
Add a button to make the editor and exercise go fuller of screen.

screenshot 2018-07-30 13 50 53

screenshot 2018-07-30 13 50 58

License: MIT
Signed-off-by: Oli Evans <oli@tableflip.io>


mikeal:
Love it, let me know when you think it is ready to merge :)

olizilla:
This is good to go @mikeal tho please give the tires a good kick. It just copies the monaco dep locally, but it's still using their pre-minified bundles and monaco's own loader to pull down the same parts as before, just from our dist directory instead from a CDN.

I've tried pretty much every combination of things and this one is the least worse arrangement. Upgrading to the latest monaco and pulling the vue integration to a local component works fine too, but the bundle size is much larger and the intellisense code suggestions are significantly more annoying in that one.

PR #61, opened 2018-10-15
fix: scope output to lesson instance, fixes #60

### LINKED ISSUE #60: Error / success message should be specific to the lesson you're in
When you fail a lesson, an error message or hint (such as "You forgot to return a result :)") is displayed in a red alert zone just above the submit button. If you can't figure it out on your own and instead go back and click thorugh to the next lesson, that same failure message is still displayed on the lesson you haven't tried yet.

mikeal:


mikeal:
CI passed, i'm not sure why it is still showing as pending.

Issue #45, opened 2018-08-04
Add url to this repo and/or discuss.ipfs.io so that people know where to post questions


diasdavid:
https://proto.school/ currently has no pointer to any of our IPFS channels, making it hard for folks to know where to ask questions.

mikeal:
related (possibly duplicate of) #42

terichadbourne:
What's the best URL to link to? Both of these issues (this one and #42) reference both GH and and discuss.ipfs.io - should they both listed (with an explanation of what inquiries go where) or just one?

mikeal:
For now, I expect that most of the issues with the tutorials will actually be with the tutorials themselves, so let's just link to this repo **for now.**

In the future we may want to link into a more general discussion repo or to the IPFS discuss. One concern I have with the IPFS discuss is that the moderation settings aren't super friendly to new people and proto school is scoped a little bit broader than IPFS.

terichadbourne:
Okay, I'll plan to add a link to https://github.com/ipfs-shipyard/proto.school/issues in a spot that's visible regardless of which specific tutorial you're working your way through.

terichadbourne:
Closed by merging of PR #63

Issue #42, opened 2018-08-02
Add link to Github repo to report issues


vmx:
Add a link to the website with a link to the Github repo and a note that users can report bugs there.

terichadbourne:
See discussion in duplicate issue #45. Will mark both issues closed once I've added a link.

terichadbourne:
Closed by merging of PR #63

Issue #18, opened 2018-07-26
Domain and deployment


ghost:
Just quick questions for keeping track of infra things:

- Is this deployed through IPFS or github pages?
- Where's the domain registered? (We should move it to Gandi and DNSimple)

mikeal:
* github pages
* iwantmyname.com, i can migrate it to whatever we like whenever we decide to do so.

mikeal:
For the DNS migration, let's do that after DWeb but before we do any kind of public promotion about this resources.

lgierth:
Thanks! Yeah let's definitely do it only *after* DWeb Summit

mikeal:
It's after DWeb :)

Let me know when it's a good time to migrate the DNS :)

lgierth:
Cool -- could you give me access to the account so I can just do it in one go, without coordination? If there's anything else in the account, I can also initiate a traditional domain transfer, and you'd give me the token, and later confirm the transfer.

---

You can do something like what PeerPad does: https://github.com/ipfs-shipyard/peer-pad/blob/master/ci/Jenkinsfile

The ci/Jenkinsfile triggers the build, and `record` sets a branch-recordname mapping, i.e. master commits will update /ipns/dev.peerpad.net and production commits will update /ipns/peerpad.net.

Regardless of any branch/record mappings, you get the website built and the link displayed via Github's commit build status.

mikeal:
It's my personal iwantmyname.com account so I'd prefer to just walk through the transfer in a private channel (Slack?).

> RE: Jenkins

I don't have any opinions about the deployment environment. I assume that, given I don't have access to infra, I won't be responsible for maintaining it, so whatever the people doing the work are comfortable with is fine with me. I would just prefer that:

* Master commits deploy to "production."
* Each commit/push gets an IPFS-only content-addressable deploy.

lgierth:
Okay get me a transfer authorization code :)

lgierth:
We'll have to wait until the domain is at least 60 days old: https://www.gandi.net/en/tlds/school/rules

terichadbourne:
@lgierth @mikeal Is this issue still relevant? If so, has enough time passed for us to make the change and close out this issue? 

[...]

PR #63, opened 2018-10-16
Question link

terichadbourne:
Added a message to the bottom of the lesson template that encourages users to submit their questions/feedback on the lesson, linking to a new issue with the lesson number / title / path pre-filled and a body that explains the type of feedback we'd appreciate. 

Addresses issues #45 and #42.

### DIFF
--- a/src/components/Lesson.vue
+++ b/src/components/Lesson.vue
@@ -178,11 +178,8 @@ export default {
       parsedExercise: marked(self.$attrs.exercise || ''),
       parsedConcepts: marked(self.$attrs.concepts || ''),
       lessonTitle: self.$attrs.lessonTitle,
-<<<<<<< HEAD
       issueUrl: `https://github.com/ipfs-shipyard/proto.school/issues/new?labels=question&title=Question+on+Lesson+${self.$route.path.slice(self.$route.path.lastIndexOf('/') + 1)}:+${self.$attrs.lessonTitle}+(${self.$route.path})&body=Have%20a%20question%20or%20suggestion%20regarding%20a%20ProtoSchool%20lesson%3F%20Please%20use%20this%0Atemplate%20to%20share%20it!%0A%0A1.%20URL%20of%20the%20lesson%20that's%20confusing%3A%0A%20https%3A%2F%2Fproto.school%2F%23${self.$route.path}%0A%0A2.%20What%27s%20confusing%20about%20this%20lesson%3F%0A%0A3.%20What%20additional%20context%20could%20we%20provide%20to%20help%20you%20succeed%3F%0A%0A4.%20What%20other%20feedback%20would%20you%20like%20to%20share%20about%20ProtoSchool%3F%0A`,
-=======
       output: self.output,
->>>>>>> master
       IPFS,
       expandExercise: false,
       options: {
--- a/src/components/Lesson.vue
+++ b/src/components/Lesson.vue
@@ -180,6 +180,7 @@ export default {
       parsedExercise: marked(self.$attrs.exercise || ''),
       parsedConcepts: marked(self.$attrs.concepts || ''),
       lessonTitle: self.$attrs.lessonTitle,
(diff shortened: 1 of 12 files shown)


olizilla:
@terichadbourne thanks! Pls can you drop a screenshot of what you expect to see on any PR with a visual change, so when we run it, we know that we're on the same page.

terichadbourne:
Thanks for the great feedback, @olizilla. I'll work on those updates. 

Here's the visual change expected, the addition of this sentence/link under the exercise box on every lesson page on the site: 

image

terichadbourne:
My latest commit to this branch should address all the issues that @olizilla flagged, as well as adding encoded formatting to the body of the new issue. 

@mikeal suggested that I create the `issueUrl` in the `data: self` object instead of the `computed` object, but when I try to add it to the list there as `issueUrl: <the ridiculously long string>,`
the page doesn't render at all, but also produces no errors in Terminal or the console. 

If I instead try `issueUrl: **return** <the ridiculously long string >,` it says "Failed to compile" in the browser and thinks there's an unexpected token. 

Do I need to adjust some `this` or `self` references somewhere to make it work from this section of the code? For reference, this is how the variable is used elsewhere in the code: 
```
<div>
        Feeling stuck? We'd love to hear what's confusing so we can improve
        this lesson. Please <a :href="issueUrl">share your questions and feedback</a>.
      </div>
```

(What's been committed is the working code before attempting this maneuver, not the failed maneuver.)

olizilla:
I'd just go with the `computed` property. I don't know vue, but the docs suggest that computed props are the place to things that you want to derive from other props, as we're doing here

https://vuejs.org/v2/guide/computed.html

terichadbourne:
This branch now also addresses issue #65, correcting a grammar issue in success messages.

terichadbourne:
@olizilla Would you have a minute tomorrow to test this updated `question-link` branch for me? Two new things to check for: 

1. You should now see the success message "Everything works!" in most lessons (previously "All works!") . _addresses issue #65_
image

2. In the final lesson of each series (basics 3 and blog 7), when you correctly solve the problem you should see the success message "Great job! You've completed this series of lessons!" and a "More Workshops" button that takes you back to the list of workshops, rather than a "Next" button that takes you to a non-existent subsequent lesson (blank white screen). _addresses issues #64 & #38_
image

src/components/Lesson.vue
```diff
@@ -95,6 +98,12 @@
         </div>
       </div>
     </section>
+    <section class="indent-1 mb4 mw-900">
+      <div>
+        <p>Feeling stuck? We'd love to hear what's confusing so we can improve
+        this lesson. Please <a :href="issueUrl">share your questions and feedback</a>.</p>
```
olizilla: This would be more friendly if it opens github in a new tab, as we don't currently preserve your state across page reloads... right now if you click on it, your answer is lost.

```html
<a :href="issueUrl" target="_blank">
```

src/lessons/Basics/03.vue
```diff
@@ -37,7 +38,7 @@ const validate = async (result, ipfs) => {
   }
 
   if (result.value === 1 && result.remainderPath === '') {
-    return {success: 'Great job!'}
+    return {success: "Great job! You've completed this series of lessons!"}
```
olizilla: minor, but we're using singlequotes `'` for strings in this project. The bigger issue is we should have CI checks and code linting as part of the PR process, but that's an issue for another PR!

src/lessons/Blog/07.vue
```diff
@@ -170,7 +171,7 @@ const validate = async (result, ipfs) => {
   } catch (err) {
     return {fail: `Your function threw an error: ${err}.`}
   }
-  return {success: 'All works!'}
+  return {success: "Great job! You've completed this series of lessons!"}
```
olizilla: single quotes again

terichadbourne:
Great suggestions, @olizilla! I'll fix those items. I really appreciate you clueing me in to the nuances of formatting choices we're trying to be consistent with in the code across the repo.

terichadbourne:
Actually, @olizilla, I just remembered why I used double quotes there, and it's because that string includes an apostrophe in `you've`.  Would you rather I put backticks (template string format) instead of double quotes on the outside of strings when that's the case?

olizilla:
@terichadbourne I don't think we have anything set up to auto deploy this code yet, so it won't appear on the live site until it's manually updated. There is an issue to fix that #18 and another to fix the CI testing setup #71

olizilla:
@mikeal can you take a look at the CI/CD stuff when you get well? also, get well sir!

PR #86, opened 2018-11-19
[WIP] feat: save lesson code state, fixes #13

### LINKED ISSUE #13: Save State
Currently, the back and forwards buttons don't work and nothing is saved on restart. It would be nice if we could save state per lesson (just record the last code autosaved from the editor).

mikeal:


terichadbourne:
@mikeal This isn't ready to merge, but could you take a look at the current state after my commits? It's totally functional, but I had to manually update the `cachedCode` binary from within other functions because otherwise Vue didn't notice the changes. I think I've worked around those issues there, but I'm encountering similar problems as I tackle the saving of state in terms of whether a lesson has been completed, so I'd love for you to see if there's some concept about how and where Vue recognizes changes that I'm missing. I'd love feedback on whether there's a more efficient way to use Vue than what I'm doing.

src/components/Lesson.vue
```diff
@@ -258,11 +262,27 @@ export default {
...
+        this.code = this.$attrs.code || defaultCode
+        this.editor.setValue(this.code)
+      }
+    },
     onMounted: function (editor) {
       this.editor = editor
     },
     onCodeChange: function (editor) {
-      // console.log(editor.getValue())
+      if (editor.getValue() === this.$attrs.code) {
```
mikeal: `this.$attrs.code` probably needs to be `(this.$attrs.code || defaultCode)` to handle the case where the code property wasn't set by the lesson.

src/components/Lesson.vue
```diff
@@ -258,11 +262,27 @@ export default {
...
+        this.clearCache()
+      } else {
+        let cacheKey = this.$attrs.exercise + this.$attrs.lessonTitle
+        localStorage[cacheKey] = editor.getValue()
+        this.code = editor.getValue()
+        this.cachedCode = true
```
mikeal: Perhaps we can set cachedCode to some kind of string so that we can have different text when the code is loaded from cache and when we've *just* started saving it?

terichadbourne:
Dropping a couple of links here for myself on how to get Vue to "watch" localStorage, since it's not working properly yet.  
- Vue documentation on client-side storage: https://vuejs.org/v2/cookbook/client-side-storage.html
- Other options from StackOverflow: https://stackoverflow.com/questions/42974170/is-there-any-way-to-watch-for-localstorage-in-vuejs

src/components/Lesson.vue
terichadbourne: @mikeal Since I'm evaluating, not setting, then what you're proposing would mean that it evaluates to true if the code in the editor matches either thing, right? If `this.$attrs.code` exists, then I only want it to evaluate to true if it matches that, not `defaultCode`.   If `this.$attrs.code` exists but the editor matches `defaultCode` instead, I need it to evaluate to false. Do I need to something more like this? 

`if ((this.$attrs.code && editor.getValue() === this.$attrs.code) || (!this.$attrs.code && editor.getValue() === defaultCode))`

mikeal: The simplest way is to just do an inline OR operator in the assignment:

```javascript
if (editor.getValue() === (this.$attrs.code || defaultCode) )
```

terichadbourne: Ah, okay, so putting the parens around those options will force it to match the first if the first exists. Thanks!

terichadbourne:
@mikeal I thing we've hit MVP+ on this update. There are some design tweaks I'd like to make still, but I'd love to merge this now so I can share the functionality in my video update. Do you have time to take a quick look and give it a thumbs up?

src/components/Lesson.vue
```diff
@@ -79,14 +94,14 @@
...
+          Update the code to complete the exercise. Click <strong>submit</strong> to check your answer.
           </div>
         </div>
         <div class="pt3 ph2 tr">
-          <div v-if="output.test && output.test.success && lessonNumber === lessonsInWorkshop">
+          <div v-if="((output.test && output.test.success) || lessonPassed) && lessonNumber === lessonsInWorkshop">
             <Button v-bind:click="workshopMenu" class="bg-aqua white">More Workshops</Button>
           </div>
-          <div v-else-if="output.test && output.test.success">
+          <div v-else-if="(output.test && output.test.success) && lessonPassed || lessonPassed">
```
mikeal: you might want to put one more set of parens in here cause it's not clear to me how the conditions are being paired.

terichadbourne: Good call. Was actually able to simplify this one.

src/components/Lesson.vue
terichadbourne: Done!

src/components/Lesson.vue
```diff
@@ -224,10 +245,22 @@ export default {
...
     this.output = {}
+
+    // doesn't work to set lessonPassed in here because it can't recognize lessonKey yet
+  },
+  updated: function () {
+    // runs on page load AND every keystroke in editor AND submit
+    // console.log('in UPDATED and this.cachedCode is: ', this.cachedCode)
+  },
+  beforeUpdate: function () {
+    // console.log('before update')
```
mikeal: we should probably remove the commented out debugging logs :)

src/components/Lesson.vue
```diff
@@ -258,11 +304,66 @@ export default {
...
+      // runs on page load, NOT on every keystroke in editor
+      // console.log('in MOUNTED and localStorage[this.lessonKey] is: ', localStorage[this.lessonKey])
+      // console.log('this.lessonPassed is: ', this.lessonPassed)
+      // console.log('in MOUNTED and localStorage[this.cacheKey] is: ', localStorage[this.cacheKey])
+      // console.log('this.cachedCode is: ', this.cachedCode)
+      // don't need to update lessonPassed here because it sets itself correctly
+      // in 'data' on page load
       this.editor = editor
+      if (this.cachedCode) {
+        console.log('PRESUMPTION: Returned to a lesson previously cached (1)')
```
mikeal: From this point on down there are a lot of debugging statements still int he code.

terichadbourne:
Screenshots of results: 

image

image

Issue #91, opened 2018-11-30
New Tutorial: Mutable File System


terichadbourne:
Following up on a Slack convo with @mikeal  (cc @olizilla & @marcooliveira based on recent related convos)

The existing ProtoSchool workshops (basics and blog) use the DAG API, but we never introduce it as one of many APIs for working with IPFS. I had incorrectly thought that the data being stored in IPFS in these examples just happened to be JavaScript objects because it was a simple example to demonstrate, not realizing that it's the only type of data that can be stored with that particular API and that you need to use a different API to use IPFS for file transfer. 

Since file transfer is a more practical use case that a beginner could see a need for, it would be nice to also build a simple lesson using the file API.  We could either build a net new beginner workshop using the file API or replace the existing basics workshop. My inclination is to do the former and make it clear that one workshop is about transferring data and one about transferring files, linking between them so users can see the alternatives. We might also want to link from this lesson about transferring files yourself to the GUI file sharing service at  share.ipfs.io so users have an example of the types of applications they could build using this API.

The blog workshop would remain using the DAG API, since it's a more advanced lesson about structured data that makes sense in that format.

Note that this is *not* being proposed as part of the 1.0 release.

mikeal:
The thing I was stuck on mentally for a while was "how do we get a file into the browser based code workflow." I thought that we'd have to figure out a way to incorporate UI programming into the lesson.

The other day I thought of a much better solution. We create a new Lesson sub-component called "FileLesson" and in the exercise box we have a drop area for files. Instead of a run function that takes no arguments the run function will take the arguments of the files dropped into the drop area.

The first lesson should pass automatically if they drop anything in, just to get them accustomed to the workflow. The next lesson or two should just be a quick primer on working with File objects in the browser, cause the API is pretty rough around the edges.

Then we can get into how to store files in IPFS :)

terichadbourne:
Would the lesson not be possible with them typing filepaths (`my/computer/awesome.doc`) because there's some lack of ability to access their file system from the editor, or is it just friendlier the way you're describing @mikeal?

mikeal:
@terichadbourne not unless we have them download an Electron desktop app or something. The browser security model won't allow us any direct file access without the user explicitly dropping the file into our web application.

mikeal:
Alternatively, instead of a "drop area" we could have a button that asks them to select a file. I usually implement a drop area that also kicks off that workflow if you click it instead of dropping a file in.

[...]

PR #111, opened 2019-01-03
file upload functionality and error hijacking

mikeal:
Still a work in progress, but the basics are there.


mikeal:
I implemented this as a mixin but I'm not so sure that's the best approach. Most of the code still has to live in the main template and code eval anyway, so we could just add the functionality to Lesson and just have the lesson author add the "isFileLesson" property.

mikeal:
Also, I'm not sure how we should handle directories (if at all). There's a way to do it but I think it's Chrome only https://github.com/feross/drag-drop/blob/master/index.js#L121

lidel:
Oli did a deep dive on this recently and  I think we have code that works across Firefox/Chrome in [ipfs-webui](https://github.com/ipfs-shipyard/ipfs-webui). 

@olizilla are you able to link to relevant snippets?

terichadbourne:
I do think we'll want to demonstrate some things using directories, so tips from @olizilla on the implementation @lidel mentioned would be great.

lidel:
In case it helps, I tracked some prior art:
- file streaming: [ipfs-companion/2.7.0/add-on/src/popup/quick-upload.js](https://github.com/ipfs-shipyard/ipfs-companion/blob/v2.7.0/add-on/src/popup/quick-upload.js) 
  (`pull-file-reader` solves chrome crash with bigger files, described in https://github.com/ipfs-shipyard/ipfs-companion/issues/464)
- directory uploads and working drag&drop: [ipfs-webui/v2.3.3/src/bundles/files.js](https://github.com/ipfs-shipyard/ipfs-webui/blob/v2.3.3/src/bundles/files.js)
  (includes various fixed related to directory upload mentioned in https://github.com/ipfs-shipyard/ipfs-webui/pull/931#issuecomment-453486161)


Digression: directories complicate things, initial lesson may want to deal with uploading files alone. Later, we could show how to create directories on the fly (https://github.com/ProtoSchool/protoschool.github.io/issues/91#issuecomment-460976703), and finally we could tackle directory uploads.

terichadbourne:
@mikeal I've pushed the changes we worked on today to 

TODO: Troubleshoot why upload click is still making the file selector open twice in a row. (Drag and drop functions as intended.)
TODO: Make the validation happen when the submit button is clicked, NOT when the file is uploaded. (Flow should be that the user uploads files, updates code editor, then hits submit to test.)
TODO: Turn `uploadedFiles` from a binary into an array of file names uploaded and print them to page in place of the placeholder content.

terichadbourne:
Made additional UI improvements based on some of the feedback from today's design team and GUI team calls. 

Suggestions to consider that aren't yet implemented: 
- Make the new file icon aqua like the add icon (I don't know much about SVGs and this one is put there in a different way from the other so I need to study up on how to do that, but it seems like an easy one to knock off)
- Replace "Reset Code" and "Start Over" text links with some sort of consistent reset icon 
- Disable submit button until both steps one and two are completed OR gray out and disable the code editor until the file upload is complete. (I don't love the latter idea since it doesn't really matter in what order you do the steps.)
- Separate instructions for step 1 and instructions for step 2 (this would require a pretty decent refactor of how our markdown files and Vue files interact

terichadbourne:
@alanshaw Any chance you could help get me on track tomorrow? I think I've built a kinda-almost-reasonable lesson on `files.add()` and submitted successful code in the editor but can't access what I mean to to do the validation. And I'm definitely confused by the `add` documentation and could use some straightening out to improve my descriptions. 

Feedback from anyone is welcome. This is how you'd find the files that produce http://localhost:8080/#/file-basics/02 (in case you haven't played with Vue or this project before): 

image

mikeal:
I talked with Teri in a sideband. The issue is `files.add()` vs `files.write()` which is sort of a recurring theme. I’m starting to wonder if we should hide the non-mfs API’s in a namespace somewhere.

-Mikeal

________________________________

alanshaw:
@terichadbourne there is no `ipfs.files.add()`, you'll need to use `ipfs.files.write()` for MFS. Non MFS APIs are currently at the root level `ipfs.add()`, `ipfs.get()` etc.

I'm sure you've been referencing them already but just incase, docs are here: https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/FILES.md

Could you expand on what's confusing about the add documentation?

mikeal:
@alanshaw the main thing I find confusing for new users is that the API’s that only encode data into IPFS file terms also ask for paths to those files. The default assumption of the user is that those files are then referencable by the paths they used when adding the files, but of course they aren’t stored in MFS.

If we want to keep the encoding APIs around and think there are good reasons for people to use them then it would help to add a code sample to `.add()` on how to reteive that data after you’ve stored it using `.get()` and the CID. We should probably do this for all of the APIs that write to IPFS since we have quite a few of them across different namespaces.

terichadbourne:
@alanshaw Regarding what I found confusing... 

I did go into this knowing that add() was not part of MFS, but spaced on the fact that it would be an issue when the validation method suggested by @mikeal (files.ls()) was MFS. And I think I wrote ipfs.files.add() by accident but it still seemed to work (proven by some console logging, not my MFS-based test despite the documentation just suggesting ipfs.add(). 

Here was my mindset when I went to figure out how to add a file to IPFS. I have a file. The way I know to refer to it is by its name and path on my computer. I went to the documentation looking for where I would plug the path into the function, and I found these options for the formats you can pass in: 
```
ipfs.add(data, [options], [callback])
Where data may be:

a Buffer instance
a Readable Stream
a Pull Stream
an array of objects, each of the form:
{
  path: '/tmp/myfile.txt', // The file path
  content: <data> // A Buffer, Readable Stream or Pull Stream with the contents of the file
}
```
and I saw "file path" and got excited. But as I eventually figured out that's an imaginary/virtual/whatnot destination filepath, not a source filepath. The examples were stuff like this: 

```
const files = [
  {
    path: '/tmp/myfile.txt',
    content:  ipfs.types.Buffer.from('ABC')
  }
]
``` 
in which I eventually realized the contents of that `Buffer.from()` are a text string that's the content of a document. If my goal is to upload a file and it knows what its own contents are, why am I supposed to rewrite the contents in a string? (This is me still thinking the path was a source path. Can you tell I haven't buffered before? 😂) 

I pinged Mikeal and he suggested that the file I was working with (one uploaded to the browser already) might already be the right kind of format to just pass in alone, which turned out to work. I ended up just doing `ipfs.add(file)` (accidentally `ipfs.files.add(file)`) where `file` was the object already stored in the browser and it worked. But I have absolutely no idea which of the format options above I was matching, so I can't write good lesson content around it, as is obvious [here](https://github.com/ProtoSchool/protoschool.github.io/blob/9071b2cff2f2564472c71dbad8d1b00f411513ac/src/lessons/File-Basics/02.md). 

If I'm using the non-MFS `add()`, would I only ever use this `path` option when I use the `wrapWithDirectory: true` that you told me about on our call?

alanshaw:
> If we want to keep the encoding APIs around and think there are good reasons for people to use them

The plan is not to, we want to merge the files APIs together into something that is a lot less confusing for users to get to grips with, but this is a little way down the line, although there's no reason you can't get involved in the discussion now https://github.com/ipfs/interface-js-ipfs-core/issues/284 - a proposal for a united API would be rad!

> it would help to add a code sample to `.add()` on how to reteive that data after you’ve stored it using `.get()` and the CID. We should probably do this for all of the APIs that write to IPFS since we have quite a few of them across different namespaces.

In the mean time we need to make the documentation as clear and approachable as possible. This is good feedback - do either of you have time to help action this?

alanshaw:
> If I'm using the non-MFS `add()`, would I only ever use this `path` option when I use the `wrapWithDirectory: true` that you told me about on our call?

I think the key thing to understand with `ipfs.add()` is that `path` is the path you want the `content` to be accessed at _after_ it is added to IPFS.

`wrapWithDirectory` just ensures everything you add goes in the same directory and so it's accessible by a single CID. You don't have to `wrapWithDirectory`, but if you want everything you add accessible from a single CID + path (`/ipfs/QmHash/*`) you _either_ have to specify `wrapWithDirectory: true` or make sure all your paths begin with the same directory name:

e.g.

```js
[
  { path: '/root/a.txt', content: Buffer.from('a') },
  { path: '/root/b.txt', content: Buffer.from('b') }
]
```

...and then you get back the CID for "a", CID for "b" and the CID of the directory "root".

`wrapWithDirectory: true` adds a root directory to all your paths for you automatically.

As a side note, JS IPFS doesn't support adding content with multiple root directories yet so you can add multiple items with no path (`''`) and get lots of individual CIDs back but you can't currently add `/root0/a.txt` and `/root1/b.txt` in the same shot (unless you specify `wrapWithDirectory: true`).

mikeal:
@terichadbourne one of two things is happening:

* The IPFS implementation isn’t being set properly.
* The createNode() API was removed.

You should be able to edit https://github.com/ProtoSchool/protoschool.github.io/blob/0c73dd8cb1c27be08d0a44fcb4eba812161d963f/src/components/Lesson.vue#L307 and change from using `IPFS.createNode()` to `new IPFS()`. If it still fails with the same error then it isn’t being set properly.

terichadbourne:
@mikeal If I change that line to: 
`return new IPFS({repo: Math.random().toString()})`
the error message on submit changes to: 
`Lesson.vue?99a6:351 Uncaught (in promise) TypeError: IPFS is not a constructor at eval (Lesson.vue?99a6:351)`


The left side here is the `code` branch (live website, which doesn't error out) and the right side is the `file-upload` branch from this PR. The only differences I see are the `args` on lines 315 and 333 on the right. I checked and every lesson (not just the new file upload ones) in this new branch refuse to validate, which would suggest the problem is at the component level and not necessarily with the Vue files for the new lessons. On the live site, validation is still working. 

image

Does any of this provide helpful clues?

terichadbourne:
@mikeal If I do this: 
```
    createIPFS: function () {
      if (this.$attrs.createIPFS) {
        return this.$attrs.createIPFS()
      } else {
        let ipfs = this.IPFSPromise.then(IPFS => {
          console.log("inside createIPFS and IPFS is: " + IPFS)
          return IPFS.createNode({repo: Math.random().toString()})
        })
        return ipfs
      }
    },
```
it errors out at the console log like so: 

`Lesson.vue?99a6:351 Uncaught (in promise) TypeError: Cannot convert object to primitive value at eval (Lesson.vue?99a6:351)`

terichadbourne:
Successfully fixed that error by updating dependencies with Mikeal's help. 🎉

terichadbourne:
I'm feeling pretty stuck, having made drafts of both a [lesson 3](http://localhost:8080/#/file-basics/03) meant to be about `ipfs.add()` and a [lesson 4](http://localhost:8080/#/file-basics/04) meant to be about `ipfs.files.write()`. There are questions embedded in the text of each lesson because I don't thoroughly understand the docs, and neither lesson has working validation, again because I don't thoroughly understand the docs. (I am not proposing we actually write these lessons back to back, I'm just trying to figure out how I'd do the exercises and validation and explain the concepts. )

My understanding is that @mikeal doesn't think we should teach non-MFS commands at all but that @alanshaw and most of GUI & IWB team think it's necessary, which is why my [proposed outline](https://github.com/ProtoSchool/protoschool.github.io/issues/91#issuecomment-456457377) produces two separate tutorials and covers both sets of actions. I know @alanshaw intends to use some of this content for IPFS Camp so I want to make sure we cover the needed areas. 

I'd really appreciate it if someone familiar with the APIs (both normal and MFS) could sit down with me now that this is in progress to take a look at what I'm misunderstanding in the docs (which will probably be obvious from reviewing my lesson text or validation code) and help me get unblocked on the validation for these lessons. 🙏

@mikeal With [lesson 3](http://localhost:8080/#/file-basics/03) about `ipfs.add()`, the user (even when submitting code I think is a correct solution) is seeing an error message that's different from the ones I provided for the lesson: 
`Error: invalid input: expected buffer, readable stream, pull stream, object or array of objects`
I've searched the whole project and can't find that string. Is that being generated by the IPFS API itself and somehow surfaced here? Does the internal value returned inside the `run` function (in this case, the `addedFiles` array) have to fit one of a number of formats for the overall validation to work? Like only CIDs are acceptable return values across all lessons? 

cc @fsdiogo

mikeal:
> My understanding is that @mikeal doesn't think we should teach non-MFS commands at all

I should clarify. I don’t think we shouldn’t teach them **at all**, but that they should be advanced topics taught *after* mfs.

A new developer, with no prior experience or understanding of IPFS, will have certain expectations about something calling itself a “file system.” A simple operation like “write this file to this name” should then show up in a list of files you’re storing under the name you gave it. This is a fundamental aspect of all filesystems developers have ever worked with, and it’s more or less exactly how mfs works.

If we want to move onto advanced topics from that point on it should be to go from MFS, to an explanation of how the files and directories are encoded, to the non-MFS files API. Teaching things in the opposite direction only makes sense to people who understand how IPFS is actually implemented.

terichadbourne:
Thanks for clarifying your take on this, @mikeal. 

Do you still envision teaching the DAG API first? The reason the current order made sense to me is: You've first learned some simple add/get commands from the DAG API (and no one has ever presented IPFS to you as a "file system" yet - it feels more like shoving something into a database to me, and I don't see anything on ipfs.io that suggests it's a file system) and then you learn how to do the same commands for files in the non-MFS way. Then you raise MFS as "Oh, but there are all these other things we normally love to do with files... I really just want Dropbox. Wouldn't it be cool if we could do that too?" 

In my mind (with an admittedly limited understanding), MFS is a fake layer of abstraction covering the truth of how IPFS works, and I think there's some knowledge of IPNS required to understand how it's working to make you feel like you have a more stable system. It feels slightly disingenuous to me to teach it as though it's a real/simple thing and then reveal that it's not, but I'm open to other perspectives here.

mikeal:
DAG API should be an advanced topic, for sure. Adding files to mfs is very easy and a great introduction to the stack and to decentralized technology. The DAG API is a bit deeper, which was kind of clear to us already because we built a tutorial on “decentralized data structures” as a pre-requisite to teaching it.

-Mikeal

________________________________

mikeal:
MFS is quite useful without IPNS and doesn’t necessarily require it. You only need IPNS if you want to share a mutable endpoint with other peers, but if you think about js-ipfs like any other browser library, or go-ipfs as any other service you run locally, they are addressable *locally* for development in the way most developers would expect.

Sure, you’re teaching the stack from the top layer of abstraction on down, but that’s how we teach most things. In fact, most web developers don’t learn how HTTP really works until they are pretty far along in their journey as a developer. Only weird old people like me learned TCP/IP and protocols from the ground up :P

I’d also keep in mind that if what developers want to do is add files to a browser node, list them reliably and share the CID’s with other peers to download p2p, the basic MFS tutorial effectively accomplishes that. You don’t need to know how the files are constructed as data structures. If you look at similar libraries like WebTorrent, they expose some high level APIs and 90% of the users get by with just that.

-Mikeal

________________________________

terichadbourne:
> You only need IPNS if you want to share a mutable endpoint with other peers

@mikeal I'm probably missing something super obvious here, but if I don't care about sharing with peers, why would I learn to use MFS on IPFS instead of just using a file system I'm familiar with? 

> The DAG API is a bit deeper, which was kind of clear to us already because we built a tutorial on “decentralized data structures” as a pre-requisite to teaching it.

Do you not envision that “decentralized data structures” tutorial being a prerequisite to the MFS tutorial as well?  Maybe the merkle stuff is irrelevant but the content addressing and trust stuff is, right?

terichadbourne:
@mikeal I think I have mostly working code now for Lesson 4 on `files.write`, but I wanted to make an error message for if someone gives the root directory and forgets to add a file name to it. I built that in like this: 
```} else if (!itemsAreFiles) {
    return {'fail': 'Uh oh. It looks like you created a folder instead of a file. Did you forget to include a filename in your path?'}
```
But instead when I submit code that seems to be wrong in the intended way, it never even gets to my validation code and instead I get this: 
image

Is there away around that so I can give a more useful error message?

Also wanted to create an error for when someone forgets the `{create: true}` but this auto-error is confusing since "file" in the error !== `file` in my code: 
image
I only confirmed that was what this meant by doing this: 
image

Am I pretty much stuck with the existing error messages that I assume are coming from the Files API or can I circumvent this to provide something better for my situation?

terichadbourne:
@alanshaw Heads up that I logged this issue in the js-ipfs repo after @mikeal tried to help me fix my Lesson 3 (`ipfs.add(data)`). It seems from the docs like you can't use blobs there as `data` and that's what the file object in my browser is, so we're not aware of a simple way to get the uploaded-to-browser file into IPFS without using MFS which supports blobs in the browser.  Are we on track with this or mistaken about the functionality? 

https://github.com/ipfs/js-ipfs/issues/1972

@fsdiogo Does share.ipfs.io use only MFS or was a workaround found to use the other methods as well?

mikeal:
@terichadbourne ya, these are all come from the Files API. We have two options, we could re-write a bunch of the eval code to pass exceptions to the validation function, or we could make these IPFS errors a bit nicer and more self-explanatory.

fsdiogo:
> @fsdiogo Does share.ipfs.io use only MFS or was a workaround found to use the other methods as well?

[share.ipfs.io](https://share.ipfs.io) does not uses MFS at all, it's just plain simple adding and getting!

terichadbourne:
@olizilla tells me that the way the GUI team has been making Web UI work with browser file objects is by wrapping them in readable streams, and of course I'd rather avoid having to teach ProtoSchool users how to do that. @hugomrdias has it in his OKRs for the upcoming quarter to make the non-MFS add() work with blobs. 🎉

@olizilla is on the same page as @mikeal about teaching MFS first, and given that I can push ahead on that with blob support now, my plan is to focus this PR on an MFS tutorial and later circle back to see what we can effectively teach on the non-MFS commands in the browser.

@alanshaw I know you're depending on some of this content for your IFPS Camp tutorial, but I'm hoping this prioritization will be okay. 🤞

terichadbourne:
I've restructured this tutorial to be just about MFS, have removed the old `ipfs.add` lesson, and have added an `ipfs.ls()` lesson. 

I'd love feedback from @fsdiogo @mikeal or others on whether the content of the existing lessons is actually true and whether I can simplify any of the existing validation. Good tips now will help me be more efficient moving forward. Thanks!

terichadbourne:
@fsdiogo Heads up that I've just added content for the next lesson (files.mkdir) but haven't yet fixed up the validation.

fsdiogo:
@terichadbourne @olizilla can you please take a look at the error overriding and the instructions for using it, more or less in https://github.com/ProtoSchool/protoschool.github.io/pull/111/commits/6a2f2ee7f36d2bda7060faedc40750e5a6474b9f and https://github.com/ProtoSchool/protoschool.github.io/pull/111/commits/68e8e3646ddd8b92f118f9281284e5622575a990?

@terichadbourne as this branch is getting a bunch of work done in it, we should fork it to different branches so our work won't collapse, but specially to ease the code reviews.

I've commented the imports to the MFS lesson in my last commit, so we can merge this branch to `code`. You should then fork `code` and continue your work on the MFS lessons; I'll fork code too, to add the console outputs in another branch; @olizilla will make his own branch too from `code` and get on with his work as well.

Sounds good?

fsdiogo:
Ping @terichadbourne @olizilla ☝️

terichadbourne:
@fsdiogo I’m all for your plan for getting this merged and splitting out work off into new branches. I should be able to make some time to proof this today, but I’d prefer if someone more familiar with IPFS than I am also proofed the new validation system for hijacking error messages. @mikeal or @olizilla could you please help to review?

terichadbourne:
@fsdiogo I've clarified and added to your instructions for using the new error override feature. 

The end result is working is for me as a user when I uncomment the routing to get the MFS lessons usable - I see the custom errors when needed but can make native syntax errors and such pass through unimpeded.

terichadbourne:
@fsdiogo, @Mikeal just gave the official “LGTM” after looking at your commits for the error hijacking. Feel free to merge if you’re comfortable with my latest commit.

.travis.yml
```diff
@@ -7,6 +7,7 @@ cache:
 before_script:
   npm install
 script:
+  npm test
```
olizilla: @fsdiogo needs more hypens:

```yaml
script:
  - npm test
  - npm run build
```

PR #121, opened 2019-01-08
Make 'prev: null' valid for treePostCid (fixes #43)

### LINKED ISSUE #43: Blog lesson 6 can be solved differently
Blog lesson 6 "Chain of links" could also be solved with having `prev: null` on the oldest blog post. It's a correct solution, but it won't validate at the moment. Make it validate.

terichadbourne:
- Make 'prev: null' a valid option for the treePostCid (previously
treePostCid could not have a 'prev' field at all)

### DIFF
--- a/src/lessons/Blog/06.vue
+++ b/src/lessons/Blog/06.vue
@@ -121,6 +121,9 @@ const validate = async (result, ipfs) => {
   const dogPostCid = 'zdpuAxe3g8XBLrqbp3NrjaiBLTrXjJ3SJymePGutsRRMrhAKS'
   const computerPostCid = 'zdpuAwwT4kGJxT7mgVZRgvmV3ke8qGNZGLuCgLhJsdBSQGM44'
   const treePostCid = 'zdpuAri55PR9iW239ahcbnfkFU2TVyD5iLmqEFmwY634KZAJV'
+  const treePostCidPrevNull = 'zdpuAoNUinwYTMoTR8Wq7945MKSSpAUNGW1d1wkTHhRcchG3D'
+  const computerPostCidWhenTreePostCidPrevNull = 'zdpuAsFHXZkpXcjuERjACPp1pAs9J7b4cdYtn9Dv9xBcAGhWV'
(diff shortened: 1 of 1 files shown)


terichadbourne:
@vmx Could you please confirm that this successfully addresses your issue #43 by making it a valid response to add 'prev: null' to the 'treePostCid' instead of not adding any 'prev' field all? 

I accomplished this by submitting the 'prev:null' solution, copying the hashes out of the resulting error messages and pasting them in as correct options in the validation section. (Presumably the better long-term solution would be to refactor the code for Lesson.vue so it knows how to take multiple potential solutions, but that's a bit beyond my grasp at present.)

src/lessons/Blog/06.vue
```diff
@@ -140,25 +140,25 @@ const validate = async (result, ipfs) => {
...
+    return {fail: `The "computers" blog post should link to the "trees" blog post, but it links to ${treeNode.content}.`}
   }
-  if ('prev' in treeNode) {
+  if (('prev' in treeNode) && (treeNode.prev !== null)) {
     return {fail: 'The "trees" blog post shouldn\'t link to other blog posts.'}
   }
 
   const computerNodePrevCid = computerNodePrev.toBaseEncodedString()
-  if (computerNodePrevCid !== treePostCid) {
+  if ((computerNodePrevCid !== treePostCid) && (computerNodePrevCid !== 'zdpuAoNUinwYTMoTR8Wq7945MKSSpAUNGW1d1wkTHhRcchG3D')) {
```
vmx: I guess that hash is the treePostCid when `prev: null`. I think it would be clearer if it would also be a variable, something like `treePostCidPrevNull` or so. This way people know where that hash came from.

If you want to have something (un)equal to two items, you could write it also as `[treePostCid, treePostCidPrevNull].includes(computerNodePrevCid)`. I'm unsure if that's clearer or not, I just wanted to mention it.

This whole comment also applies to the hashes below.

terichadbourne: Both are great suggestions, thanks!

terichadbourne:
@vmx I've made both updates you suggested. Would you have a moment to pull the changes and run again on your machine to double-check that everything's working properly before we merge? Thanks!

Issue #108, opened 2018-12-19
Improve responsive design, esp nav bar


terichadbourne:
The navigation bar currently gets cut off on narrower screen widths, so we need a hamburger menu or the like to make it accessible on mobile. 

Other site content works better responsively, but we should do a quick review of it when fixing the nav to see what else needs improvement.

terichadbourne:
Currently on a phone or other narrow screen, the nav bar extends further right than the viewport, while the rest of the content stays in the viewport. 

![image](https://user-images.githubusercontent.com/19171465/50597828-bceba380-0e76-11e9-8559-80da64b2d014.png)

If you scroll right to see it, the background is missing, so if you visit a page and its title gets highlighted in white, it's invisible. For example, while on the "Host" page: 


![image](https://user-images.githubusercontent.com/19171465/50597862-dee52600-0e76-11e9-9c1d-a09489277078.png)

mikeal:
In addition to just making things viewable on mobile, we need to have a deeper conversation about what our longer strategy is for mobile as the current experience w/ the code editor is not something that will work on mobile.

olizilla:
There is the super helpful components section to the tachyons docs which shows various patterns for nav bars and lots of other layouts:

http://tachyons.io/components/#nav

The simplest thing here might be to just let the nav scroll horizontally like

http://tachyons.io/components/nav/list-overflow/index.html#

zebateira:
We can close this @terichadbourne given this was tackled in #123 ?

terichadbourne:
Yeah, we can call it good. If you use responsive setting on a laptop there's a width at which the nav will create a horizontal scroll before flipping to the hamburger, but when checking the phone views in dev tools it's showing up consistently as a hamburger, so this is probably too much of an edge case to worry about.

PR #123, opened 2019-01-10
Make nav responsive - fixes #122

### LINKED ISSUE #122: Bug: nav menu isn’t responsive on mobile
I tried protoschool on my iPhone 10 in Chrome and noticed some issues with responsive wrap - especially with the navigation menu at the top not responsively collapsing into a hamburger or something. I don’t believe this is device specific, but would show up on all sufficiently narrow mobile devices, but can give more detail if others can’t reproduce. Either a “more” Carey with a drop down menu for the items that don’t fit or a hamburger style would allow those menu items to be accessible without the weird scrolling.

terichadbourne:
WORK IN PROGRESS

Working to create a responsive nav, ideally a hamburger menu, to avoid the overflow issue noted in issues #108 & #122.

Edits made so far fix the overflow but require scrolling rather than using a hamburger menu.


terichadbourne:
I've figured out one option for changing from `flex-direction: row` to `flex-direction: column` in order to stack the nav items on a narrow screen, but I don't know the ideal way to make it flip over at the ideal screen width. (The media query I'm using doesn't seem like the ideal Tachyons plan, but the Tachyons breakpoints aren't appropriate.) 

image

The other issue is that there are two versions of the nav, one for tutorial pages and one for other pages. The one that loads on tutorial pages acts more like breadcrumbs to help orient the user to where they are in the content, and it looks very silly in the current vertical layout (see below), so we may need to do some adjusting for the two versions.
image

Obviously we also need to add a hamburger icon and X icon to hide/display this vertical nav. I'm envisioning that sitting at the right side of the navy portion at the top of the screen, to the right of the ProtoSchool logo, but I'm open to suggestions. Note that the navy portion lives in the `Header.vue` file while the actual nav lives in the `Navigation.vue` file.

terichadbourne:
On medium screen widths in case of overflow, a scroll bar now enables access to all options in the horizontal nav. 

In small screen widths (as judged by Tachyons standard) a mobile nav appears showing the current page name and the hamburger icon: 

image

Or breadcrumbs showing tutorial name if in a tutorial: 

image

The mobile nav expands into a vertical section and goes away when an option is selected or the "X" icon is clicked: 

image

terichadbourne:
@olizilla Thanks so much for the pairing today! I've think I've managed to accomplish the additional steps we discussed when we parted ways. Any chance you'd have time to quickly review this PR your Monday morning so I can get it merged before the launch?

terichadbourne:
I've merged to ensure the new functionality is live as we have more eyes on the site today during launch. If anyone spots an issue, please let me know. Thanks!

olizilla:
@terichadbourne just tried it on my phone and it works well! The only problem with it is the hamburger button needs more padding as it's difficult to tap reliably on phone screen at its current size.

Much more minor: on small screens, the header and the navbar have a smaller horizontal padding than the main area text does. It would look neat if they lined up, at least at the left edge.

terichadbourne:
Thanks @olizilla! I was wondering about the padding issue and knew if I added it to just the hamburger icon it would give inconsistent padding with that main area. Opened a new issue #128 to make sure I circle back and fix this.

PR #127, opened 2019-01-14
feat: add chapter info for london

olizilla:
http://github.com/protoschool/london is alive


terichadbourne:
Looks great! Thanks, @olizilla!

PR #133, opened 2019-01-15
feat: code split ipfs module using import()

mikeal:
This PR lazy loads the IPFS module in the lesson component. This should result in a significantly smaller bundle size for the index page of the site.


terichadbourne:
Just worked through a few lessons on localhost and don't see any new issues emerging. Merging!

Issue #130, opened 2019-01-15
License


agentofuser:
What is the license for the content currently on the website?

Are new tutorial contributions supposed to be made under a specific license? Will there be any kind of copyright assignment?

mikeal:
Everything should be (Apache2 + MIT) and we should make that clear throughout the project. There's no CLA or copyright assignment as part of the contribution process.

terichadbourne:
@mikeal I see Apache 2.0 or MIT separately under license options when you create a new repo. Is there a combined version we're using in other projects that you could point me to? I'm happy to update our current repos to add the license.

mikeal:
You just put both licenses in the LICENSE file and note "Apache-2.0 AND MIT" as the license in the package.json licenses field :)

PR #135, opened 2019-01-16
Add license 

terichadbourne:
Partially addresses #130, although we may need to apply licenses to other repos as well. 

@mikeal I'm not at all positive I've correctly interpreted your suggestion in #130. 

I don't know how "There's no CLA or copyright assignment as part of the contribution process." fits into what we add to our page, or how it fits with the fact that both of the proposed licenses include a place for a copyright assignment.

 I've filled in the copyright assignment in the licenses as Protocol Labs. Is this correct, and is it right that it's 2019 (should it be 2018-2019 based on when work started)? 

I"ll get your feedback on this PR first before trying to apply a license to other repos. Do we need one on the organizing repo and the individual chapter repos as well?

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,29 @@
+# License
+
+ProtoSchool is licensed under both the Apache 2.0 and MIT licenses, as noted below.
+
+## Apache License, Version 2.0
+
+Copyright 2019 Protocol Labs
+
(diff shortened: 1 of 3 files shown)


terichadbourne:
closes #130 (opened new issue for organizing repo)

PR #137, opened 2019-01-20
docs(project): Added Mumbai chapter details

indcoder:


terichadbourne:
Thanks for the PR @indcoder! A few quick requests for consistency between listings: 

- Is there a commonly-used abbreviation we could use for the state of Maharashtra to keep fairly consistent formatting in the listings?  In the US we're using _Boston, MA_ and _Denver, CO_ for example. (Or better yet, if there is only one Mumbai in India, the state or province could be excluded, as has been done with London, Porto, etc.)
- Could you please change "India" to "IN" as we're using each country's [ISO-3166 2-letter country code](https://en.wikipedia.org/wiki/ISO_3166-2) for consistency?
- Would you mind changing the region to just "Asia"? I'm happy to reassess when we have more chapters, but for now I'm envisioning full continents as the best differentiator here so we don't end up with too many region names. 

Thanks!

indcoder:
Hi @terichadbourne ,

1. Yes we have an 2 letter abbreviation, I will shorten it to that.
2. I will but I took a cue from the US based cities in the json file; they are not following s ISO-3166 2-letter country code

image.

3. Will do.. Any reason why the field should not be renamed to continent itself?

terichadbourne:
Thanks for the updates @indcoder! I actually caught and fixed the US abbreviations earlier today. Just merged this so your chapter should be listed on the site within a few minutes. Thanks again for organizing!

PR #174, opened 2019-03-26
Fixed some eslint errors

andrew:
Reduces the number of errors that show up when running `npm run serve` for a nicer contributor experience.


terichadbourne:
Thank you @andrew!

PR #189, opened 2019-04-08
chore: add navbar link to landing

fsdiogo:
Multiple times I've found myself wanting to click the marked link to go to the tutorial landing page, but that wasn't a link.

nav

✅ It is now!

### DIFF
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -109,6 +109,7 @@ export default {
 .nav-link {
   margin: 2px 20px;
   font-size: 18px;
+  font-weight: 700;
   text-decoration: none;
   cursor: pointer;
 }
--- a/src/components/Navigation.vue
+++ b/src/components/Navigation.vue
@@ -12,10 +12,7 @@
         <!-- standard nav  -->
         <div v-else class="dn flex overflow-auto items-center bg-aqua white pv3 center tc mw7">
           <div v-for="link in links">
-            <router-link v-if="link.path === $route.path"
-              class="nav-link white" :to="`${link.path}`">{{link.text}}</router-link>
-              <router-link v-else
-              class="nav-link navy" :to="`${link.path}`">{{link.text}}</router-link>
+            <router-link :class="[isActive(link) ? 'white' : 'navy ', 'nav-link']" :to="`${link.path}`">{{link.text}}</router-link>
           </div>
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


terichadbourne:
@fsdiogo Great idea!

New issue when I load the site from this branch and shrink the width of my window: 

image

Compared to the current version of the site which looks like this at narrow widths: 
image

Can you please ensure the fonts match between "Tutorials" and the tutorial shortname, and ensure the space before the ">" doesn't disappear? 

I had previously used the "fake-nav-link" class to make the tutorial shortname white, following the convention that something you are already viewing is highlighted in white, as in: 

image

What are your thoughts on making the tutorial shortname white again? I agree with you that if you click it you should go to the landing page, which isn't where you currently are if within a lesson, but should you still have that visual indicator that you're in the right place? 

Interestingly, in your current implementation if I get to the landing page from a link on the tutorials index, the header looks like this: 
image

But if I then click "Data Structures" in that nav, which just takes me to where I already am, it turns white and looks like this:

image

terichadbourne:
@fsdiogo In the same vein as the previous question, should Tutorials be white in this dropdown? 

image

fsdiogo:
> In the same vein as the previous question, should Tutorials be white in this dropdown?

Yeah it should, that's not the case because the URL doesn't match `tutorials`.

Should the lessons URLs be in the form of `/tutorials/basics/01`, etc?

terichadbourne:
@fsdiogo I presume we can fake the highlighting in the same way I did it previously with that fake nav link. Just check if you're on a lesson page and make the Tutorials line of the dropdown appear to be selected.

Are you asking whether we should change the path for the various tutorials to add `/tutorials/`? We didn't do that originally because tutorials are the main content of the website and we didn't anticipate any clashes between tutorial shortnames and other pages on the website (eg there won't be a tutorial with shortname `chapters` or `build`). I'm not totally opposed to adjusting this if you can make a good case for it, but for sharing purposes I consider shorter URLs to be preferable.

fsdiogo:
> @fsdiogo I presume we can fake the highlighting in the same way I did it previously with that fake nav link. Just check if you're on a lesson page and make the Tutorials line of the dropdown appear to be selected.

Great, thanks! This is now taken care of.

> Are you asking whether we should change the path for the various tutorials to add /tutorials/? We didn't do that originally because tutorials are the main content of the website and we didn't anticipate any clashes between tutorial shortnames and other pages on the website (eg there won't be a tutorial with shortname chapters or build). I'm not totally opposed to adjusting this if you can make a good case for it, but for sharing purposes I consider shorter URLs to be preferable.

I was, but yeah, I don't have a strong opinion about this so let's keep it this way.


PS: Ready for review.

terichadbourne:
@fsdiogo Tutorials is now highlighted in the hamburger menu, but I still see a font mismatch here in mobile between "Tutorials >" and "Tutorial Shortname." Inspector says the former is taking a font weight of 500 and font size of 1rem from the div above it (Tachyons fw5 and f5) and latter is taking font size of 14px from .nav-link within the media query and font weight of 700 from from the plain old .nav-link. The former also can't change to white and be clickable. 

I seem to be able to resolve this locally by changing `link` to `nav-link` on line 26 in Navigation.vue: 
        `<router-link class="nav-link navy" to="/tutorials">Tutorials</router-link>`

(I can commit and push if you want but my package-lock.json has a bunch of local commits I'd rather not push.) 

image




This issue I noted earlier still seems to be the case in your current implementation: 

 _if I get to the landing page from a link on the tutorials index, the header looks like this:_ 
image

_But if I then click "Data Structures" in that nav, which just takes me to where I already am, it turns white and looks like this:_

image

Do we care enough to fix that discrepancy?

fsdiogo:
> (I can commit and push if you want but my package-lock.json has a bunch of local commits I'd rather not push.)

I've fixed, thanks for the heads up.

> Do we care enough to fix that discrepancy?

Fixed!

terichadbourne:
@fsdiogo Looking good!

PR #192, opened 2019-04-09
[WIP] Chore/add social meta tags

### LINKED ISSUE #191: Add meta tags to support social cards
I'd like to ensure that when people link to proto.school from social media platforms, a card is produced. 

Currently the Twitter card validator can't come up with anything at all to use: 
image

Ideally we could ensure support for all of Twitter, Facebook, LinkedIn, etc. by using Twitter and Open Graph meta tags in our head.

Some resources:

https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started
https://developers.facebook.com/docs/sharing/webmasters/
https://css-tricks.com/essential-meta-tags-social-media/

At the moment, the whole site is parsed as a single page, so for the time being my guess is that even if you provided a link to a specific URL, all URLs would have to use the same card as each other.  So our first step is to get sharing working well for the site as a whole, and when we tackle #179 to help crawlers see the site as separate pages, we can try to find a way to implement separate meta tags for each page.

terichadbourne:
closes #191 

@fsdiogo Do you know what the real URL would be for an image living in src/static/images? Or whether there's somewhere else I'd need to put an image to be referenced within a metatag for social? 

Also, I accidentally checked out my new branch from your branch instead of code. 🤦‍♀️


terichadbourne:
My best interpretation of https://cli.vuejs.org/guide/html-and-static-assets.html#static-assets-handling led me to move that image to the `public` folder. 🤷‍♀️

fsdiogo:
> @fsdiogo Do you know what the real URL would be for an image living in src/static/images?

Those images pass through webpack and have an hash appended to them, they're supposed to be used with relative paths, as said in the link you posted.

> My best interpretation of cli.vuejs.org/guide/html-and-static-assets.html#static-assets-handling led me to move that image to the public folder. 🤷‍♀️

Yes, everything you put in the `public` folder won't go through webpack and you can access them directly by `site.url/<your_file>`.

So, the simplest solution is putting it in the public folder, and it works. We should be able to reference them relatively too, and webpack would take care of it, but as it's just one single image, I think this is good to go for now.

Issue #182, opened 2019-04-02
Feature: Surface console logs into the UI


terichadbourne:
per a chat with @fsdiogo... 

Currently there are a few lessons where we ask the user to open their console and see something printed there. For example, we might ask them to view a CID to see how it's different before and after an action. It would be great if we could print console logs (or even better, only specific ones) to the screen instead so the user doesn't need to figure out how to open the console and can keep their focus in our UI.

Something similar is done in IPLD Explorer.

fsdiogo:
👋 I've been doing some experiments on how we could possibly tackle this and I've got two proposals:


**1)** Using [JSON.stringify()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) with a bit of formatting;

<img width="792" alt="stringify" src="https://user-images.githubusercontent.com/33324750/56149938-95dc5c00-5fa5-11e9-9260-fb89f2875626.png">

- Pros
  - the result is clean and readable
  - we don't have to use any other third-party lib for this
  - as we're already using `highlight.js`, I'm trying to syntax highlight this output to make it easier on the eyes and more appealing (haven't got it working, but it should be something on my end)
- Cons
  - no expand/collapse functionality like in the dev tools, so if the output is big we'll get a lot of text


**2)** Using [vue-object-view](https://github.com/ebuzek/vue-object-view);

<img width="808" alt="vue-object-view" src="https://user-images.githubusercontent.com/33324750/56149939-95dc5c00-5fa5-11e9-863b-d43b86185d5c.png">

- Pros
  - gives us the ability to expand/collapse (but it's strict and not customizable)
- Cons
  - adding a third-party lib that doesn't seem to have any activity in a while
  - no syntax highlighting (or so it seems)


I'm in favour of using **(1)** and getting the syntax highlighting working, but as I lost some time on trying **(2)** it's worth mentioning it! Haven't found any other lib like `vue-object-view`, if you know some please mention it so I can try it out!

fsdiogo:
Oh, I forgot to mention it, the way I've wired this up is that you just have to  pass a key/value to the object of the return of the `validate` function, with `log: foo` (where `foo` is what you want to log) and `logDesc: bar` (where `bar` is the description below the `Step 3: Check the data`. If those keys aren't found that step is not shown.

mikeal:
[...]

PR #198, opened 2019-04-17
feat: show logs on ui

fsdiogo:
Surfaces the console logs on to the UI, so user won't have to open their dev tools.

Closes https://github.com/ProtoSchool/protoschool.github.io/issues/182.


fsdiogo:
We should merge https://github.com/ProtoSchool/protoschool.github.io/pull/111 so people can then branch of `code` and work on features without getting blocked.

To review this you can check commit https://github.com/ProtoSchool/protoschool.github.io/pull/198/commits/23b293a0232138a74b22245ab770e90b9de2f4bf and https://github.com/ProtoSchool/protoschool.github.io/pull/198/commits/f814ce225551bc331c4641fb0d3a3ef859b9ceab, but https://github.com/ProtoSchool/protoschool.github.io/pull/111 has to be merged first.

fsdiogo:
Added syntax highlighting to the outputs:

sh

fsdiogo:
@terichadbourne this is ready for review, can you take a look at the code and the instructions for adding this?

To test you'll have to uncomment the MFS lessons and pass the Lesson 3.

terichadbourne:
@fsdiogo I don't know if it happened in this PR or earlier, but I see this in the standard lesson boilerplate: 
const code = `const CID = require('cids')

const run = async () => {
  /* your code here */
}
return run
`

Can we remove `const CID = require('cids')? The `code` const is what's shown in the exercise box and I'd like it as streamlined as possible. This also looks like something only some authors might need. How else can we get CID functionality accessible without cluttering the editor? (Note that I don't see this line in the file upload boilerplate.)

terichadbourne:
@fsdiogo I made two commits here. 

The first makes it so that it only adds a "step 3" header when it's a file upload lesson, since normal lessons don't have step 1 and step 2 labels. 

The second is a bunch of tweaks to the instructions for developing lessons in general. 

Feel free to merge this PR if my changes look okay to you!

fsdiogo:
> Can we remove const CID = require('cids')?

Yeah, we can remove it AFAIK.

> How else can we get CID functionality accessible without cluttering the editor? 

To get that functionality we only need to pass the modules to the `data`, so we don't need to clutter `code` with that.

```js
const modules = { cids: require('cids') }

export default {
  components: {
    Lesson
  },
  data: () => {
    return { text, validate, code, modules }
  }
}
</script>
```

PR #199, opened 2019-04-18
fix: travis config

fsdiogo:
Removes `npm run test` as we don't have any test for now and was breaking the CI for the lack of `-` when running multiple scripts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ cache:
   directories:
     - node_modules
 before_script:
-  npm install
+  - npm install
 script:
-  npm run build
+  - npm run build
 notifications:
   email: false
 deploy:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,7 +7,6 @@ cache:
 before_script:
   npm install
 script:
-  npm test
   npm run build
```
olizilla: Please add hyphes for all the things so that we're less likely to hit the issue again.
```suggestion
  - npm run build
```
same for the `before_script`. Notice how travis shows all there recommended configs with hypenated lists for all the things, even lists of 1: https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#ember-apps

PR #200, opened 2019-04-22
feat: add MFS tutorial

terichadbourne:
- Uncomment MFS lessons in `main.js`
- Add MFS tutorial to courses in `courses.json`


terichadbourne:
@fsdiogo I'm having a problem with your new UI logging feature as i work on lesson 4. I just adjusted the message for when you forget `{ long: true}`. If I forget to include that and submit the code, it shows the correct `logDesc` and `log`. However, if I then add the { long: true } and hit submit again, only the `logDesc updates`, not the ` log`. 

if I reset code and paste in the same correct answer with `{ long: true }` included, both the `logDesc` and `log` are correct.

fsdiogo:
@terichadbourne just fixed the https://github.com/ProtoSchool/protoschool.github.io/pull/200#issuecomment-485523432 issue in PR https://github.com/ProtoSchool/protoschool.github.io/pull/205.

fsdiogo:
Closed this one by accident, sorry. Already synced with the code highlighting fix from `code`.

terichadbourne:
Heads up that @olizilla and I have reworked the proposed layout for lesson order in the issue: https://github.com/ProtoSchool/protoschool.github.io/issues/91#issuecomment-487084184

terichadbourne:
@fsdiogo @olizilla Can either of you easily figure out what's broken in MFS lesson 3 (the new stat lesson)? I'm getting 
```
Syntax Error: SyntaxError: /Users/terichadbourne/code/pl/protoschool/proto.school/src/tutorials/Mutable-File-System/03.vue: 'import' and 'export' may only appear at the top level (50:0)

  48 | const modules = { cids: require('cids') }
  49 | 
> 50 | export default {
     | ^
  51 |   components: {
  52 |     Lesson
  53 |   },
```
But I had gone back to the boilerplate (for Lesson, not FileLesson) and copied the bottom part so I'm not clear what's going wrong here.

fsdiogo:
@terichadbourne yup, you're missing an `}` on line 30 😄

terichadbourne:
Thanks @fsdiogo! 🤦‍♀️

terichadbourne:
@fsdiogo I appear to need help with your logging tool. In the MFS tutorial, lesson 3 (the one that lets you practice uploading and see what data is now available for each file in the browser) used to make you open your inspector to see a console log of `files` and return files.length. The console log looked like this: 

image

I want to instead have the user return `files` and then use your logging tool to show what it contains, but it keeps printing as an array of objects with no values in them.  

```
[
   {},
   {},
   {}
]
```

although it console logs like the previous thing did: 

image

I've tried using `result`, `JSON.stringify(result)`, `JSON.stringify(result, null, 2)`, etc. but they all come out blank. I'm at about an hour of futile console logging. Do you know what's going wrong here? 

UPDATE: I looped through all the objects in the array to make new objects like so: 
```
  const filesArray = []
  const files = result
  for (var i = 0; i < files.length; i++){
    var file = {
        'lastModified': files[i].lastModified,
        'lastModifiedDate': files[i].lastModifiedDate,
        'name': files[i].name,
        'size': files[i].size,
        'type': files[i].type,
        'webkitRelativePath':  files[i].webkitRelativePath
    }
    filesArray.push(file)
  }
```

It now works if I do `log: JSON.stringify(filesArray, null, 2)`  

image

but that feels like a whole lot of work. Did I really have to do all that or did I miss a simpler solution?

And are there additional instructions we should add to our instructions about stringifying when you do the logging?

terichadbourne:
@fsdiogo I'm having some trouble implementing an error override in the MFS lesson 7 for `mkdir`. The error I want to catch is if someone forgets to use the option `{parents: true}` when trying to make a nested directory. This mistake causes the built-in error "file does not exist" which I want to override: 

image

With that option used, the code passes my validation: 
image


What I tried to do was add `:overrideErrors="true"` at the top of the Vue file where I call the FileLesson template and add the following to my validation: 

```js
} else if (result && result.error.message === 'file does not exist') {
 return { fail: 'The path to the directory you\'re trying to create can\'t be found. Did you forget to use the \{ parents: true \} option?'}
```

But if I put in the bad code and hit submit, nothing happens in the UI and the following is in the console: 

image

I tried running `npm install` (and then restarting my server) in case it was something to do with that file it's not finding, but that doesn't fix the problem. Any idea what's wrong?  

Note that my last commit pushed has the broken validation code commented out so it won't break. I couldn't figure out a comment format that worked inside the template portion so that part is just emitted in my last commit.

fsdiogo:
> I've tried using result, JSON.stringify(result), JSON.stringify(result, null, 2), etc. but they all come out blank. I'm at about an hour of futile console logging. Do you know what's going wrong here?

@terichadbourne what's happening here is that what is returned (the `result`) is a [File object](https://developer.mozilla.org/en-US/docs/Web/API/File), and it appears that can't be logged as you would normally do. 

Compare it with the MFS lesson 5 (the `ls` command), everything works as expected because the result is just a plain old JS object.

That's why it worked when you just iterate through the file properties and store them in an object to log.

I'll try to find a simpler way to do this.

> And are there additional instructions we should add to our instructions about stringifying when you do the logging?

I don't think so. This is a JS quirk and not IPFS, but if you want we can mention this in the troubleshooting section on the README.

fsdiogo:
@terichadbourne, in commit https://github.com/ProtoSchool/protoschool.github.io/pull/200/commits/45cec488d6b4ff81a26bafae080b71c65fd96ecd I added an utility function that you can pass an array of files and it'll do the heavy lifting of "converting" it to a log-friendly object 🙂 

I've also refactored lesson 3 to use that function.

terichadbourne:
Thank you @fsdiogo. I'll check that out!

As I was hitting my snooze button this morning I had a thought about why I might be getting that `result.some is not a function` error when trying to override in lesson 7, so I'll take a stab at fixing that. I'll be pairing with @olizilla on some MFS tutorial validation stuff in a bit.

fsdiogo:
> As I was hitting my snooze button this morning I had a thought about why I might be getting that result.some is not a function error when trying to override in lesson 7, so I'll take a stab at fixing that.

That's happening because your `contains()` function is getting called in the ifs conditionals and breaking because `result` is not an array, it's an error object.

One way to fix that is to check if `result` is an error object in the initial ifs, and then get on your way or not!

We can do some pair-programming on that if you need to.

terichadbourne:
Thanks @fsdiogo - yeah, that was what I figured out, so I changed the order the code runs in and added a conditional to stop it from getting into the place where it gets stuck. I appear to have it fixed now, but I'll let you know if I get off track again.

terichadbourne:
@fsdiogo Hoping you can figure out why my error override isn't working. I would have sworn it was yesterday. 🤦‍♀️

If you do MFS lesson 7 but forget `{parents:true}` in your `mkdir` (so use this to reproduce): 
```js
  await Promise.all(files.map(f => ipfs.files.write('/' + f.name, f, {create: true})))
  await ipfs.files.mkdir('/some/stuff')
  let rootDirectoryContents = await ipfs.files.ls('/', {long: true})
  return rootDirectoryContents
}
return run
```

IPFS returns a "file does not exist" error: 

image

which I thought I had overridden here: 
image

but instead the user gets no feedback from our own logging tool and the native error also isn't returned.

Could you help me figure out what's happening?

terichadbourne:
@fsdiogo @olizilla @mikeal  I discovered an interesting quirk while making the `ipfs.files.mv` lesson (lesson 8) that I'm hoping one of you can explain. 

This is what I've ended up showing as the correct solution for moving all the files (only files, not directories) from the root directory into the `/some/stuff` directory: 
```js
/* global ipfs */
  await Promise.all(files.map(f => ipfs.files.write('/' + f.name, f, {create: true})))
  await ipfs.files.mkdir('/some/stuff', { parents: true })
  let rootDirectoryContents = await ipfs.files.ls('/', {long: true})

  rootDirectoryContents.forEach(item => {
    if (item.type === 0) {
      ipfs.files.mv('/' + item.name, '/some/stuff')
    }
  })

  let someStuffDirectoryContents = await ipfs.files.ls('/some/stuff', {long: true})
  return someStuffDirectoryContents
}
return run
```

But what I had assumed I'd need to do, mostly because I've had to do it with every other IPFS command I've ever used, was this (adding` await`): 

```js
/* global ipfs */
  await Promise.all(files.map(f => ipfs.files.write('/' + f.name, f, {create: true})))
  await ipfs.files.mkdir('/some/stuff', { parents: true })
  let rootDirectoryContents = await ipfs.files.ls('/', {long: true})

  rootDirectoryContents.forEach(item => {
    if (item.type === 0) {
      await ipfs.files.mv('/' + item.name, '/some/stuff')
    }
  })

  let someStuffDirectoryContents = await ipfs.files.ls('/some/stuff', {long: true})
  return someStuffDirectoryContents
}
return run
```
If I do that, it fails and logs this to the console: 
```
08.vue?3dd8:33 result:  {error: SyntaxError: await is only valid in async function
    at new Function (<anonymous>)
    at _callee…}

15:46:09.484 vue.runtime.esm.js?2b0e:619 [Vue warn]: Error in v-on handler (Promise/async): "Error: /some/stuff does not exist"

found in

---> <Button> at src/components/Button.vue
       <FileLesson> at src/components/File-Lesson.vue
         <Anonymous>
           <App> at src/App.vue
             <Root>
warn @ vue.runtime.esm.js?2b0e:619
logError @ vue.runtime.esm.js?2b0e:1884
globalHandleError @ vue.runtime.esm.js?2b0e:1879
handleError @ vue.runtime.esm.js?2b0e:1839
(anonymous) @ vue.runtime.esm.js?2b0e:1856
Promise.catch (async)
invokeWithErrorHandling @ vue.runtime.esm.js?2b0e:1856
invoker @ vue.runtime.esm.js?2b0e:2179
original._wrapper @ vue.runtime.esm.js?2b0e:6911

15:46:09.485 vue.runtime.esm.js?2b0e:1888 Error: /some/stuff does not exist
    at eval (index.min.js?1f76:1)
    at eval (index.min.js?1f76:1)
    at eval (index.min.js?1f76:1)
    at eval (index.min.js?1f76:1)
    at eval (index.min.js?1f76:1)
    at eval (index.min.js?1f76:1)
    at eval (index.min.js?1f76:1)
    at eval (index.min.js?1f76:1)
    at y (index.min.js?1f76:1)
    at eval (index.min.js?1f76:1)
```

I have evidence from other console logging that`/some/stuff` does indeed exist, so that part is particularly confusing. 

I can certainly remove the `await` from my examples, but could someone help me understand why I need it in all the other IPFS commands and not here?

terichadbourne:
After confirming what `result.error.message` was via logging, I tried adding this to override the error but it's not working for me. Neither the initial error nor my override are hitting the UI; the code is just erroring out and not showing any feedback to the user. 

```js
  } else if (result.error && result.error.message === 'await is only valid in async function'){
    return { fail: 'Oops! You can\'t use `await` with `files.mv` because it\'s not an async function.'}
 }
```

terichadbourne:
Update: I got my error override working and stopped the code from bailing out by fixing a scoping issue elsewhere in my code. However, I'd still love an explanation of why `await` doesn't make sense here.

terichadbourne:
In trying to build lesson 9, where the user needs to copy a file from IPFS into their MFS directory, clicking submit isn't producing any visible result to the user and the only thing in the console is that the swarm is listening. Could this just be stalled out because the public gateway is slow? 

Through logging I can tell that the run function is still trying to do this step: 

The command I'm using in my code is: 
`ipfs.files.cp('ipfs/Qme1zmi8dxBiVM7K9y5J3oPxiWWBgzA7n9M6tkmkz8kSwV', '/some/stuff/success.txt')`
(I've also tried it without providing the filename at the end of the path, and with and without `await`, all with the same results.)

I can successfully access the requested file at: https://gateway.ipfs.io/ipfs/Qme1zmi8dxBiVM7K9y5J3oPxiWWBgzA7n9M6tkmkz8kSwV 

Am I wrong about what the solution to this challenge should be or is this some kind of lag time issue? If the latter, is there any way to speed it up? Or is copying from IPFS just not viable here?

Commit with lesson 9 content is: https://github.com/ProtoSchool/protoschool.github.io/pull/200/commits/312c2434502406e3d2ab84337a9deddc865d77c9

fsdiogo:
I've taken a look at https://github.com/ProtoSchool/protoschool.github.io/pull/200#issuecomment-490241325 and the problem is the same I mentioned in https://github.com/ProtoSchool/protoschool.github.io/pull/200#issuecomment-489103204, there are errors before the validation that aren't being caught.

For example, `line 37` errors with your code:

https://github.com/ProtoSchool/protoschool.github.io/blob/3314cbf831de3d4744063592bc7a3da90f363a58/src/tutorials/Mutable-File-System/07.vue#L31-L43

I think you can solve this by changing the order of some of your functions and do more check if the result is an error or an array.

We can pair on this later if you're feeling stuck.

olizilla:
```
SyntaxError: await is only valid in async function
```
is a pretty good error message in this case... the problem is arrow functions make it easy to forget that you're in the scope of a new function. The fix for that error is to declare the function as async as below

```js
  rootDirectoryContents.forEach(async (item) => {
    if (item.type === 0) {
      await ipfs.files.mv('/' + item.name, '/some/stuff')
    }
  })
```

**but** we want to wait for all the mv calls to be completed before moving on to the `ipfs.files.ls` line, so we probably want to wait for all the promises to complete like

```js
  const files = rootDirectoryContents.filter(item => item.type === 0)
  await Promise.all(files.map(file => {
      return ipfs.files.mv('/' + file.name, '/some/stuff')
  })
```

terichadbourne:
@olizilla Thanks for the explanation!

That idea works if I change the proposed variable name from `files` to `filesToMove` to avoid a conflict and add a missing parens: 
```js
  const filesToMove = rootDirectoryContents.filter(item => item.type === 0)
  await Promise.all(filesToMove.map(file => {
      return ipfs.files.mv('/' + file.name, '/some/stuff')
  }))
```
However, as someone who was very recently very new to JavaScript, there's a lot of scary stuff in there. Giving people an easy way to use `forEach` instead of `map` or `filter` and avoiding `Promise` altogether would be cool. Here's one alternative solution I've come up with, where you pass in an array so you're making a single call to files.mv, and then you don't have to wrap `await`around a bigger function (I happened to use `map` and `filter` but wouldn't have to be comfortable with those: 

```js
  const filesToMove = rootDirectoryContents.filter(file => file.type === 0)
  const filepathsToMove = filesToMove.map(file => '/' + file.name)
  await ipfs.files.mv(filepathsToMove, '/some/stuff')
```

I've provided some examples before the exercise that show that both `'file1.txt', 'file2.txt'` and `['file1.txt', 'file2.txt']` are acceptable, so I could call that out as a clue. 

What are your thoughts, @olizilla & @fsdiogo? Can you think of an even more beginner-friendly solution?

fsdiogo:
™

terichadbourne:
@olizilla As you suggested, I took a stab at adding a text file containing the string `You did it!` in the non-MFS part of IPFS after creating IPFS (in hopes of a faster return time on the `cp` command than if I just store the file on Cluster for everyone to get), but I can't figure out how to make the add command work properly to interpret the buffer. See this commit for what I tried: https://github.com/ProtoSchool/protoschool.github.io/pull/200/commits/ce71d339e4d1e122fd4f51fab497255044897d6e

The [documentation for `ipfs.add`](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/FILES.md#add) says I'm supposed to do this: 

```
In the browser, assuming ipfs = new Ipfs(...):

const content = Ipfs.Buffer.from('ABC')
const results = await ipfs.add(content)
const hash = results[0].hash // "Qm...WW"
Now ipfs.io/ipfs/Qm...WW returns the "ABC" string.

```

but  `Ipfs` doesn't exist and it also doesn't work if I use `ipfs` or `IPFS` or exclude that whole reference and just use `Buffer`. I've also tried this unsuccessfully using the other format where you pass in both a path and content instead of just content.

```
{
  path: '/tmp/myfile.txt', // The file path
  content: <data> // A Buffer, Readable Stream or Pull Stream with the contents of the file
}
```

Could you help me get onto the course you were suggesting?

The good news is that I did successfully pass a `createTestFile` attribute from 09.vue that can be identified from within the `run` function in Lesson.vue to only run this new code conditionally. 🎉

olizilla:
it's fiddly. The only place that the IPFS constructor exists is this line:
- https://github.com/ProtoSchool/protoschool.github.io/blob/ce71d339e4d1e122fd4f51fab497255044897d6e/src/components/Lesson.vue#L388

There you have `IPFS` in scope, with all the things you'd need on it, but it's in a function that is used to construct a single ipfs instance and return it, so  it's not ideal.

Options are...  refactor code so that the IPFS constructor gets stashed somewhere you can fetch it from later, or move the ipfs dependency to not be dynamically imported (use `import IPFS from 'ipfs'` at the top of the lesson, rather than `import('ipfs')`. I guess

```js
let ipfs = this.IPFSPromise.then(IPFS => {
  // keep the constructor around so we can grab things like Buffer from it later
  this.IPFS = IPFS
  return new IPFS({repo: Math.random().toString()})
})
```

could work for now, but it feels like we should tidy up that logic pretty soon.

fsdiogo:
@terichadbourne @olizilla finished the lessons that were missing validation.

Their validation isn't as extensive as @terichadbourne standard (😄) but I think it's enough for a test run. 

When you have the time to take a look at this, please do!

src/tutorials/Mutable-File-System/02.md
```diff
@@ -1,5 +1,26 @@
...
+* **hash** (a string with the cryptographic hash)
+* **size** (an integer with the file or folder size in Bytes)
+* **cumulativeSize** (an integer with the size of the DAGNodes making up the file in Bytes)
+* **type** (a string that can be either `directory` or `file`)
+* **blocks** (if type is `directory`, this is the number of files in the directory; if type is `file`, it's the number of blocks that make up the file)
+* **withLocality** (a boolean to indicate if locality information are present)
+* **local** (a boolean to indicate if the queried dag is fully present locally)
+* **sizeLocal** (an integer indicating the cumulative size of the data present locally)
+
+It's important to note that you can `stat` your IPFS node even when you don't have anything in it yet. Even empty nodes have CIDs (hashes) that represent their contents.
```
alanshaw:
```suggestion
It's important to note that you can `stat` your IPFS node even when you don't have anything in it yet. Even empty nodes have a CID (hash).
```

src/tutorials/Mutable-File-System/02.md
```diff
@@ -1,5 +1,26 @@
...
 
-To practice, let's upload one or more files from your computer and create a console log to see what's been received by the browser as the `files` array.
+As mentioned previously, methods associated with the Mutable File System are part of the Files API, so they'll take the format of `ipfs.files.someMethod()`. Let's take a look at a simple method you can start using even before you've added any files to your IPFS node.
+
+## Exploring your IPFS node with `ipfs.files.stat`
+When working with your IPFS node, you'll often want to check the status of a file or directory. You can do this with `ipfs.files.stat`, passing in the path you'd like to check on.
+
+For example, to check the status of a directory called `stuff` located within our root directory ( `/` ), we could call the method like so:
+
+`await ipfs.files.stat('/stuff/')`
```
alanshaw:
```suggestion
`await ipfs.files.stat('/stuff')`
```

src/tutorials/Mutable-File-System/03-exercise.md
```diff
@@ -1,3 +1 @@
-Let's add some files to IPFS using MFS. Since `files` is an array representing the files currently available to us in the browser, we'll need to loop through the array and use `files.write()` to add each file we find there to IPFS, one at a time. (Only have a single file uploaded? No problem!)
-
-Put your files in the root directory (`/`) and be sure to add the name of each file to the path when you add it (**hint:** the file object in the browser stores the filename as `file.name`). Be sure to set up your options so that a new file is created when one isn't found at the given path.
+First, upload one or more files by dragging and dropping below or clicking to make a selection from your file explorer.  Next, in the code editor, remove the comment markers (`//`) that precede `return files` within the `run` function.
```
alanshaw: Screenshot 2019-05-31 at 14 40 34

Multiple file uploads do not grow the container to accomodate

src/tutorials/Mutable-File-System/05.vue
```diff
@@ -31,59 +31,66 @@ const validate = async (result, ipfs) => {
...
 }
 
 const code = `const run = async (files) => {
-  // this code adds your uploaded files to your root directory in IPFS
-  await Promise.all(files.map(f => ipfs.files.write('/' + f.name, f, {create: true})))
-  // add your code to create a new directory here
-  let rootContents = await ipfs.files.ls('/', {long: true})
-  return rootContents
+  // this code adds your uploaded files to IPFS
+  await Promise.all(files.map(f => ipfs.files.write('/' + f.name, f, { create: true })))
```
alanshaw: It's a bit odd to see this after we just did it with a loop. Why not just have the loop here again?

src/tutorials/Mutable-File-System/06.md
```diff
@@ -0,0 +1,15 @@
+As you learned in the [Decentralized Data Structures tutorial](https://proto.school/#/data-structures), CIDs (Content Identifiers) are uniquely matched to the content they represent through cryptographic hashing. Two data files with identical contents have identical CIDs (hashes) and two data files with even the smallest difference between them have distinct CIDs. The same is true for folders. Every time you update the contents of a file or directory, its CID changes.
```
alanshaw:
```suggestion
As you learned in the [Decentralized Data Structures tutorial](https://proto.school/#/data-structures), CIDs (Content Identifiers) are uniquely matched to the content they represent through cryptographic hashing. Two files with identical contents have identical CIDs (hashes) and two files with even the smallest difference between them have distinct CIDs. The same is true for folders. Every time you update the contents of a file or directory, it's CID changes.
```

src/tutorials/Mutable-File-System/05.md
```diff
@@ -1,17 +1,32 @@
...
-`await ipfs.files.mkdir('/my/beautiful/directory', { parents: true })`
+The method will default to listing the contents of your root directory ( '/' ), or
+you can choose to specify a specific `path` (directory) you'd like to inspect,
+such as `/catPics`,
+
+`file.ls()` produces an array of objects, one for each file or directory
+contained in the directory you're inspecting, with the following properties:
+
+- `name` (default): the file's name
+- `type` (optional with `{ long: true }`): the object's type (`directory` or `file`)
```
alanshaw:
```suggestion
- `type` (optional with `{ long: true }`): the object's type (`0` - file or `1` - directory)
```

src/tutorials/Mutable-File-System/07.vue
```diff
@@ -0,0 +1,187 @@
+<template>
+  <FileLesson
+    :text="text"
+    :code="code"
+    :validate="validate"
+    :modules="modules"
+    :exercise="exercise"
+    :solution="solution"
+    :overrideErrors="true"
+    lessonTitle="Create a directory">
```
alanshaw: Screenshot 2019-05-31 at 15 20 05

src/tutorials/Mutable-File-System/08.md
```diff
@@ -0,0 +1,28 @@
...
+await ipfs.files.mv(['/source-file-1.txt', '/source-file-2.txt'], '/destination-directory')
+
+// move a directory into another directory
+await ipfs.files.mv('/source-directory', '/destination-directory')
+
+// overwrite the contents of a destination file with the contents of a source file
+await ipfs.files.mv('/source-file.txt', '/destination-file.txt')
+```
+
+**Note:** `files.mv` takes longer when you move larger files. If you've been working with large files for these lessons, please click "Start Over" above the file upload window and select smaller files.
```
alanshaw: Should not be true - has this been observed? It's a bug if so!

src/tutorials/Mutable-File-System/08-exercise.md
```diff
@@ -0,0 +1,11 @@
...
+
+Remember that you can pass an array into `files.mv` as the `from` value. This is useful because it allows you to run the async function only once. Be sure to use the `await` keyword so that the `mv` call completes before the contents of your `/some/stuff` directory are evaluated.
+
+When creating your array to pass in, ensure it only contains files, not directories. Remember that each object in IPFS has a `type` property that you can use to determine whether it's a file or a directory. For files it's the number `0` and for directories it's the number `1`. (Hint: Try the [`filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) or [`forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) array methods.)
+
+Remember that `files.mv` needs paths, not filenames, so you'll need to prepend each filename in your array with `/`. (Hint: Try the [`map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) or [`forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) array methods.)
```
alanshaw: I could really do with a reminder of the output of `files.ls` here - I couldn't remember if each object had a `path` or `name` property...

src/tutorials/Mutable-File-System/09.md
```diff
@@ -0,0 +1,34 @@
...
+await ipfs.files.cp(['/source-file-1.txt', '/source-file-2.txt'], '/destination-directory')
+await ipfs.files.cp('/ipfs/QmWGeRAEgtsHW3ec7U4qW2CyVy7eA2mFRVbk1nb24jFyks',
+ '/ipfs/QmWGeRAEgtsHW3jk7U4qW2CyVy7eA2mFRVbk1nb24jFyre', '/destination-directory')
+await ipfs.files.cp(['/ipfs/QmWGeRAEgtsHW3ec7U4qW2CyVy7eA2mFRVbk1nb24jFyks',
+ '/ipfs/QmWGeRAEgtsHW3jk7U4qW2CyVy7eA2mFRVbk1nb24jFyre'], '/destination-directory')
+
+// copy a directory into another directory
+await ipfs.files.cp('/source-directory', '/destination-directory')
+await ipfs.files.cp('/ipfs/QmWGeRAEgtsHW3ec7U4qW2CyVy7eA2mFRVbk1nb24jFyks', '/destination-directory')
+```
```
alanshaw: This code block is quite unwieldy - can it be split into multiple blocks?

src/tutorials/Mutable-File-System/03-exercise.md
fsdiogo: Fixed, thanks!

src/tutorials/Mutable-File-System/01.md
```diff
@@ -1,18 +1,23 @@
...
 
-We're creating your IPFS node behind the scenes so you can focus on the content of our lessons, but eventually you'll need to learn to host your own node locally. When you're ready to experiment, you can find instructions for [installing IPFS](https://docs.ipfs.io/introduction/install/) and [initializing your node](https://docs.ipfs.io/introduction/usage/) in our docs.
+## Mutable File System
+Because files in IPFS are content-addressed and immutable, they can be complicated to edit. **Mutable File System (MFS)** is a tool built into IPFS that lets you treat files like you normally would in a name-based filesystem — you can add, remove, move, and edit MFS files and have all the work of updating links and hashes taken care of for you.
```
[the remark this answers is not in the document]
fsdiogo: Agreed, we want this to be easy to understand but it must reflect what really happens. Fixed.

src/tutorials/Mutable-File-System/02.md
```diff
@@ -1,5 +1,26 @@
-As a matter of security, web browsers don't let us change files that live on your computer. Therefore you'll need to upload one or more files to the browser that you can use throughout this tutorial.
+## Working with files in ProtoSchool
+Here in our ProtoSchool tutorials, rather than using your machine itself, we're creating a new IPFS node for you in the browser each time you hit the "Submit" button in a lesson. Whenever you see `ipfs.someMethod()` in our lessons, `ipfs` is a variable that refers to your IPFS instance, also known as a node. The actions that you take only affect your own IPFS node, not nodes belonging to your peers.
```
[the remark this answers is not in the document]
fsdiogo: > I'm not sure I understand this sentence...is it referring to the difference between running a daemon on your machine and running it in the browser?

Yeah, I think @terichadbourne was trying to differentiate between having to install IPFS and run a daemon in your terminal from running an on-demand disposable one in the browser.

src/tutorials/Mutable-File-System/08.md
terichadbourne: @alanshaw I was getting no response when I submitted code for this lesson, and the thing that seemed to fix it was uploading smaller files, so I made an assumption. It's possible I misdiagnosed the problem, but it would be great if you or @fsdiogo could try completing lesson 8 with huge files and see if you can replicate the problem. Is it possible for that issue to be true in ProtoSchool but NOT on one's own locally installed IPFS for some reason?

src/tutorials/Mutable-File-System/02.md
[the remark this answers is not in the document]
terichadbourne: Yeah, so essentially you can do all of ProtoSchool without ever learning to install IPFS. So we need to call out that we're skipping a whole step... which I can't describe accurately because I'm never done it NOT in the browser. Want to propose something that sounds accurate to you while still explaining what we're doing in the background?

src/tutorials/Mutable-File-System/02.md
```diff
@@ -1,5 +1,26 @@
...
 
-Within each exercise, you'll see that you can upload from your computer either by dragging and dropping or selecting from your file explorer. If you look closely at the `run` function in the code editor, you'll notice that it now takes an argument `files`. When you upload files from your computer, we'll make sure they're passed into the function as the `files` array. As long as you don't refresh your browser, these files will remain accessible for the next lesson in the tutorial, but you'll also have the option to upload different files to work with for each lesson.
+We're creating your IPFS node behind the scenes so you can focus on the content of our lessons, but eventually you'll need to learn to host your own node locally. When you're ready to experiment, you can find instructions for [installing IPFS](https://docs.ipfs.io/introduction/install/) and [initializing your node](https://docs.ipfs.io/introduction/usage/) in our docs.
```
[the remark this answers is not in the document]
terichadbourne: I've never done it the other way, so perhaps you can help me describe it more accurately. Would it be "eventually you'll need to learn to IPFS and run a daemon in your terminal"?

src/tutorials/Mutable-File-System/02.md
terichadbourne: There are two separate issues here - one is the plural. If you prefer it without the plural then we can switch to "Even an empty node has a CID (hash)" so the subject is singular also. 

I'm assuming your objection to "that represent their contents" was that this empty node doesn't have any contents, which is precisely what I was trying to call out. I want to remind people that the CID is directly connected to the data - or lack thereof. 

Would this make sense to you? 
"It's important to note that you can `stat` your IPFS node even when you don't have anything in it yet. Even an empty node has a CID (hash) that represents its (non-existent) contents." (or "contents (or lack thereof)")

I'm open to better suggestions!

src/tutorials/Mutable-File-System/08.md
terichadbourne: Here's a screenshot from the console if I try to use some .key files and .pdf files. 
image

terichadbourne: @fsdiogo Would you be able to tell if it were the file upload stage and not the ipfs.files.mv stage causing the slowdown? Not sure if that screenshot above sheds any light.

src/tutorials/Mutable-File-System/09.md
terichadbourne: The lesson itself stays looking shorter if we do this as a code block split by comments than if we turn those comments into text and put separate code blocks under each of them. Let's hold off and see if our other testers comment on this. If we change it here then we should change it in lesson 8 as well for consistency.

src/tutorials/Mutable-File-System/02.md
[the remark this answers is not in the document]
terichadbourne: Let me know how this sounds to you @alanshaw. Does it address your concerns? 

```
Here in our ProtoSchool tutorials, we create a new IPFS node for you in the browser each time you hit the "Submit" button in a lesson. Whenever you see `ipfs.someMethod()` in our lessons, `ipfs` is a variable that refers to your IPFS instance, also known as a node. The actions that you take only affect your own IPFS node, not nodes belonging to your peers.

We're creating your IPFS node behind the scenes so you can focus on the content of our lessons, but eventually you'll need to learn to host your own node locally by installing IPFS and running a daemon in your terminal. When you're ready to experiment, you can find instructions for [installing IPFS](https://docs.ipfs.io/introduction/install/) and [initializing your node](https://docs.ipfs.io/introduction/usage/) in our docs.
```

terichadbourne:
@fsdiogo Thanks so much for getting that file copying working while I was out! I've updated the later lessons to force the user to create a name for the file when they copy it into MFS, since the point of MFS seems to be the ability to pretend we're using a name-based system and I was super annoyed during the 20 minutes today when I thought it wasn't possible. 😂  The last few lessons still don't appear to provide any validation for issues that would be caused by touching the wrong parts of the code, but I'm going to let the testers loose on it regardless. 🎉

@alanshaw I think between Diogo and I we've addressed the bulk of your concerns, but there were a couple of spots where I held back to see what other testers say. @dietrich and @ericronne have both offered to try it out and provide feedback late this week or early next week to ensure we can incorporate feedback before we head to Barcelona. Since this is the only ProtoSchool content currently planned for use at IPFS Camp, my gut feeling is that we should go ahead and publish it to the standard site before the event and use it from there, but if you feel we need any sort of camp-specific customization, please let us know.

src/static/tutorials.json
```diff
@@ -37,14 +37,20 @@
 	},
 	"mutableFileSystem": {
 		"project": "IPFS",
-		"title": "IPFS as a Mutable File System",
-		"description": "Store, edit, and share files with with the Mutable File System (MFS).",
+		"title": "IPFS: Mutable File System",
+		"description": "All data in IPFS is content-addressed and immutable, but the Mutable File System (MFS) lets you work with files and directories as though you were using a traditional name-based file system.",
```
mikeal: This could probably be reduced to just “The Mutable File System...”, the qualifying statement at the beginning is probably not necessary.

src/tutorials/Mutable-File-System/08.md
fsdiogo: Just tested this and the slowdown isn't caused by the `ipfs.files.mv`, as it happens in other MFS lessons that operates with the uploaded file.

fsdiogo:
I added a spinner when the user submits a lesson, this is something that I wanted to do for some time but as some lessons take a to validate now was the time to do it:

spin

After the submit, the button text changes to a spinner and the button get disabled until the validation ends. This is now working for all lessons.

src/tutorials/Mutable-File-System/08.md
fsdiogo: With commit https://github.com/ProtoSchool/protoschool.github.io/pull/200/commits/f5d62e8732090f2bb9b4106ca69690280351a941 the user now gets feedback that something is in the works after submitting a lesson.

terichadbourne:
Great idea with the spinner @fsdiogo! Thank you!

PR #201, opened 2019-04-22
feat: add ui logs to blog tutorial

fsdiogo:
Add UI logs to blog tutorial and rephrased some lessons to make them clearer.

### DIFF
--- a/src/tutorials/Blog/01.vue
+++ b/src/tutorials/Blog/01.vue
@@ -1,11 +1,10 @@
 <template>
-  <div class="lesson-blog-01">
-    <Lesson v-bind:text="text" v-bind:code="code"
-            :validate="validate"
-            :exercise="exercise"
-            lessonTitle="Link an author to a blog post using its CID">
-    </Lesson>
-  </div>
+  <Lesson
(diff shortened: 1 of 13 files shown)


src/tutorials/Blog/01.vue
```diff
@@ -38,16 +36,11 @@ const _solution = `
...
-    author: samCid
-  })
-  const computerPostCid = await ipfs.dag.put({
-    content: "computers",
-    author: natCid
-  })
+  const natCid = await ipfs.dag.put({ author: "Nat" })
+  const samCid = await ipfs.dag.put({ author: "Sam" })
+
+  const treePostCid = await ipfs.dag.put({ content: "trees", author: samCid })
```
terichadbourne: Although the result of the code is the same, I consider it more beginner-friendly and easier to read when you show objects with each property on its own line. Since you'll be enabling solution viewing in a later PR, let's leave this formatting (and the line below) in the way that's easier to read.

terichadbourne:
Other than that one note above, I'm okay with this being merged @fsdiogo. Thanks!

Issue #89, opened 2018-11-29
feature: view solution


mikeal:
The lesson implementations already have the solution code in them for reference. Let's find a good way to expose this to people working to solve the lessons so that if they get stuck they can see the complete solution in order to compare.

PR #202, opened 2019-04-22
feat: add ability to view solutions in lessons

fsdiogo:
sol

Closes https://github.com/ProtoSchool/protoschool.github.io/issues/89.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ Remember that you can add comments to your default code to orient the user, such
 
 #### Provide the simplest solution to your exercise
 
-`_solution` is a string property. The value you set for `_solution` in your Vue
+`solution` is a string property. The value you set for `solution` in your Vue
 file will be used to populate the code editor if the user clicks the "View
 solution" option. (We hope you'll have provided enough clues that they won't need
 to do this!)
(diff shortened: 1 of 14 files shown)


fsdiogo:
## Workflow

solution

terichadbourne:
@fsdiogo Looking good! I pushed a couple of commits to tweak some typos or confusing language I happened to come across while testing and to add instructions for including `_solution` in your Vue file. 

Are you comfortable with "View Solution" being clickable even when you're already viewing the solution, or would you rather make the option disappear or change to non-clickable "Viewing solution" text until the user types something new and the code editor contents no longer match the solution code? This former is what happens with the "Reset code" option, which disappears when the code editor contents match the default code.

fsdiogo:
We were using `_solution` because it was like a private variable and it wasn't used. I switched to `solution` to be streamlined like the `code` one.

fsdiogo:
> Are you comfortable with "View Solution" being clickable even when you're already viewing the solution, or would you rather make the option disappear or change to non-clickable "Viewing solution" text until the user types something new and the code editor contents no longer match the solution code? This former is what happens with the "Reset code" option, which disappears when the code editor contents match the default code.

I'd prefer to always have the option to click the `View Solution`. It doesn't bother me that the `Reset code` is hidden when we have the default code on the screen.

terichadbourne:
@fsdiogo Okay, I'm good to have you merge this if you're feeling ready!  I'll  change `_solution` to `solution` in my MFS tutorials

Issue #187, opened 2019-04-07
Lesson Feedback: Blog - Lesson 7 (Traverse through all posts, starting with the most recent)


Frijol:
**URL of the lesson that's confusing:**
https://proto.school/#/blog/07

**What's confusing about this lesson?**
In the same vein as https://github.com/ProtoSchool/protoschool.github.io/issues/165, this lesson requires you to know Javascript syntax for while loops and pushing to an array

**What additional context could we provide to help you succeed?**

Maybe the best way to give JS-bonus information is e.g. "Hint: you will be using [while loops](//www.w3wchools/com/js/js_loop_while.asp) and [array methods](//www.w3schools.com/jsref/jsref_obj_array.asp) for this solution!"

**What other feedback would you like to share about ProtoSchool?**

terichadbourne:
@frijol We've just merged a bunch of work on Blog lesson 7 that we believe addresses your feedback. I'll close this out, but please feel free to re-open if we haven't achieved what you were hoping for.

Frijol:
thanks!

Issue #68, opened 2018-10-22
Blog lesson 7 instructions aren't clear enough


terichadbourne:
A few issues I ran into when trying in vain to solve blog lesson 7 (https://proto.school/#/blog/07). I'll need help better understanding some context here so I can then phrase instructions and tips more clearly to make this one easier. 

1. When do you need your CID to be an instance and when don't you? I overused the `new CID(...)` that was suggested here in my first attempt, assuming it was necessary in places it wasn't. Are the things returned by `ipfs.dag.put()` already CID instances?  Is doing the `new CID` thing the opposite of doing the  `someCid.toBaseEncodedString()` thing? You have to do it to decode it?   

2. It's unclear which things in this code block are meant to be written as is and which are meant to be replaced by some other value: `let cid = new CID((await ipfs.dag.get(someCid)).value.linkName['/'])`
I could tell, because of the "some," that `someCid` was meant to be replaced, but I didn't make that connection for `linkName`.  I now believe you're supposed to replace this with the name you gave the object that contains an object in which the field name is `'/'` and the value is the `someCid.toBaseEncodedString`, which in this case would be `prev`. Maybe we should use `fieldName` or `someLinkName` or `linkFieldName` here?  Or use coloring or wrap in `<>` symbols to highlight both of the segments you're supposed to replace (in which case we should do that throughout all lessons)? (Note that there's something related about naming of links at https://proto.school/#/blog/03 )

3. I think we should give a hint about using a while loop. IMHO, less experienced developers have more experience looping through arrays where they know how to judge the number of entries and less with figuring out how to make their loop stop itself by other means. Maybe be so explicit as to say something like: 
```Hint: How do you know when you're out of links? Try using a `while` loop and checking for the presence of a `prev` field in the current object. If it exists, you're not done yet, and you'll need to reset a variable and run the loop again.```
I'd like people spending their time thinking about how to access the right part of the CID object via IPFS commands, not thinking about the logic of the code. 

[...]

PR #204, opened 2019-04-23
fix: blog lesson 7 solution

fsdiogo:


terichadbourne:
@fsdiogo I'm testing this by making some various mistakes. If I pass in the wrong post to `return traversePosts()` near the end, it tells me I need to return 3 CIDs, which is reasonable. However, if I leave it blank or pass in something that's not a CID, I get "You forgot to return a result." Can you think of a way we could identify that they passed a non-CID value into that function, as opposed to that they deleted the word "return"  (which is what they might think we mean). If not, maybe we change the message to encompass a few potential problems.

terichadbourne:
@fsdiogo Do you think it's useful to the user to view the returned vs expected CIDs (as shown below)? I wouldn't know how to use this to reverse engineer the right solution, so I might not bother displaying it. Instead I'd suggest editing the error to suggest that it's something that might happen if you mess with the contents of the file, as below. Thoughts? 

image

fsdiogo:
> @fsdiogo I'm testing this by making some various mistakes. If I pass in the wrong post to return traversePosts() near the end, it tells me I need to return 3 CIDs, which is reasonable. However, if I leave it blank or pass in something that's not a CID, I get "You forgot to return a result." Can you think of a way we could identify that they passed a non-CID value into that function, as opposed to that they deleted the word "return" (which is what they might think we mean). If not, maybe we change the message to encompass a few potential problems.

Hmm, what if we fill the `traversePosts` function with this:

```js
const traversePosts = async (cid) => {
  const result = []
  
  // Your code goes here
  
  return result
}
```

This way we're already returning a result so that won't appear, and it enforces what the exercise says that an array should be returned.

<br />

> @fsdiogo Do you think it's useful to the user to view the returned vs expected CIDs (as shown below)? I wouldn't know how to use this to reverse engineer the right solution, so I might not bother displaying it. Instead I'd suggest editing the error to suggest that it's something that might happen if you mess with the contents of the file, as below. Thoughts?

I don't find it very useful, but as it was being displayed in the message, I've added it to the log to reduce the message pollution. Yeah, I think we can go with that! I'll commit a fix.

terichadbourne:
Oh, I actually mean the return statement at the bottom where you call the function (`return traversePosts(dogPostCid)`), so I don't think that suggestion would help, unless I'm misunderstanding you. I think what happens is that even if you write your traversePosts function perfectly, if you mess up and touch other parts of the code and pass in something that's not a CID, the traversePost function spazzes out and never hits its return statement. I do think it makes sense to let the user build a function with their return statement inside the else, though, and i'd find it confusing if you provided default code that suggested the return went at the very end. 

I'm now seeing "you forgot to return a result" if I pass in no argument on line 52, but some other errors if i pass in other non-cid stuff, for example: 

image
image
image

Since the problem is caused by people messing with the wrong part of the code, maybe we could add to the error message: "Either you forgot to return a value or you passed something that wasn't a CID into your traversePosts function. Try resetting the code and be sure to only edit the traversePosts function. Be sure the function returns an array of CIDs."

fsdiogo:
> Oh, I actually mean the return statement at the bottom where you call the function (return traversePosts(dogPostCid)), so I don't think that suggestion would help, unless I'm misunderstanding you.

You're misunderstanding. Try resetting the code, press `Enter` somewhere and submit the code: you get `You forgot to return a result :)`. That's because the `traversePosts()` function isn't returning anything. If inside `traversePosts()` you return something, that error won't appear. That's what I was saying 😛

> Since the problem is caused by people messing with the wrong part of the code, maybe we could add to the error message: "Either you forgot to return a value or you passed something that wasn't a CID into your traversePosts function. Try resetting the code and be sure to only edit the traversePosts function. Be sure the function returns an array of CIDs."

Hmm yeah, I just don't like it being a super long sentence 😐

terichadbourne:
I added some little hints so that this PR can close #187. It will also close #68.

terichadbourne:
@fsdiogo You and I are identifying two separate problems from each other. One is that the `traversePosts` function doesn't include a return statement and the other is that the user touched parts of the code they shouldn't and passed something silly (or nothing) into the function when running it (the ultimate problem being that there's a return statement inside a conditional that isn't hit when you pass in a non-CID value). Can you think of any way to distinguish between those two errors on the backend so we can provide more accurate, shorter error messages?

fsdiogo:
Hmm, as we only get the `result` of the running code, I'm not seeing a way to distinguish between those two 🤔

fsdiogo:
@terichadbourne can this be merged?

The solution that is live right now isn't ideal and although this isn't perfect still, its way better than the current version.

terichadbourne:
@fsdiogo I've added some more validation for common mistakes here and updated the assignment instructions to be more clear about what to edit and what not to touch. 

I added an override for a very common error where you try to access  `await ipfs.dag.get(cid).value.prev`, where the value doesn't exist yet because the await statement didn't get wrapped in parens and have a chance to return a value.

In setting up the override I noticed that if I follow the instructions in the readme, using `(result && result.error.message === "whatever")`, it works for the error I'm trying to catch, but for anything that doesn't have a built-in error it bails out and provides no feedback to the user who clicks submit, because `result.error` is undefined: 

image

I've changed my code to do `if (result.error && result.error.message === "whatever"` instead, but if what I'm doing is the right approach then we should fix the instructions, and if not then the instructions probably need improving regardless so I can implement it successfully in the way you intended. What's your take on what's going wrong there?

fsdiogo:
> What's your take on what's going wrong there?

The thing about this error overriding is that there are a lot of variables to consider. It depends on what the dev wants to catch and what API is using.

From my POV, we can either change the readme to 

```js
} else if (result.error && result.error.message === 'No child name passed to addLink') {
  // Forgot the file name and just used a directory as the path
  return { fail: 'Uh oh. It looks like you created a folder instead of a file. Did you forget to include a filename in your path?' }
}
```

-- OR --

better yet, explain that the test case should be up to them, what matters is that they must return an object with the `fail` key.

fsdiogo:
@terichadbourne please check commit https://github.com/ProtoSchool/protoschool.github.io/pull/204/commits/f8ebe372ebdaf25e8dd7c0723e546fc1adaca7a4 and see if it's clearer now.

PR #206, opened 2019-04-24
fix: update code highlight

fsdiogo:
Fixes https://github.com/ProtoSchool/protoschool.github.io/pull/200#issuecomment-485523432.


terichadbourne:
@fsdiogo What's the best way for me to test this since the problem I was having is in my unmerged branch. Should I merge from your branch in progress into mine?

fsdiogo:
I've tested in you branch, but we should merge from code. As this is a one line change, just copy-paste the change to your branch and see that it works 😛

terichadbourne:
@fsdiogo This is working for me now, but I'm curious... why does it do color coding the first time and not the second? Is that intentional? (see below without and with { long: true } in mfs lesson 4) 

image

image

I've leave it to you to merge if you're comfortable with that outcome. (I'm not especially concerned about the colors in this context.)

fsdiogo:
@terichadbourne I've found a more up-to-date package to highlight the logs and now everything is working properly:

PR #212, opened 2019-04-26
feat: add e2e tests for interactive lessons

olizilla:
Test that protoschool lessons are working and that the solutions
are correct. Uses cypress to walk through the basics and blog
lessons.

- add [cypress](https://www.cypress.io/) for end-to-end testing in a real brower, with a helpfulfancy user-intereface for showing what the site looked like at each step

Screenshot 2019-04-30 at 16 39 41

- Updates travis.yml to get it working in CI too

Screenshot 2019-04-30 at 16 45 33

- moves dev server to run on 3000 so as not to collide with a standard IPFS install
- adds npm run scripts for
  - `npm test` - run a dev server, wait for it to be ready then run the cypress tests in headless mode
  - `npm run cy:open` - open the cypress interface locally to debug tests. Exects you to be serving protoschool from http://localhost:3000 as you would in dev mode by running `npm start`
- adds `data-cy` attributes to the lesson component so the cypress robot can find them in the DOM  to know what state the app is in. We use these insted of `id or `class` attributes so it is clearer what there purpose is, and they are less likely to get broken by UI refactors.

With this PR merged, we can let CI tell us a PR to update to the latest js-ipfs would break anything. We can use this pattern to add more tests as we create new functionality.

License: MIT
Signed-off-by: Oli Evans <oli@tableflip.io>


terichadbourne:
@olizilla This is still meant to be WIP/draft, right? Just making sure I'm ignoring you for the right reasons. :)

olizilla:
Yep, will ping when it's ready. I'm figuring out how to run the tests in CI.

olizilla:
WOO HAA! We have it. The magic runes to make our cypress tests run on travis in a recent version of chrome.

terichadbourne:
@olizilla I just pushed a commit to update the README to reflect the new port you've suggested here. 

I tried killing my normal `npm run serve` and running `npm run cy:open` instead and got this error: 
image

and this when I try to run the test: 

image

If I do `npm run serve` and `npm run cy:open` simultaneously in different Terminal windows it seems to work. Am I meant to have to do that?

Where should the processes you've described in the PR be documented? Will tutorial authors need to be familiar with how Cypress works or just maintainers who are testing and merging PRs? Anything we should add to the readme? 

The PR @fsdiogo is working on to improve the "view solution" experience will actually break this test. He's offered to help come up with a way to repair it, but I'm wondering whether we should leave this PR open awaiting those changes or merge now and update later. Thoughts?

fsdiogo:
As PR https://github.com/ProtoSchool/protoschool.github.io/pull/215 (the `View Solution` refactor) has been merged, this won't work. Tomorrow I'll take a look to find a fix for it.

fsdiogo:
I'm having a bit of trouble figuring out a way to make this work.

I was thinking of using this approach:

1. Click view solution button
2. Copy the solution code inside the monaco editor
3. Paste the code to the editable monaco editor
4. Click the submit button and see if it passes

There are a bunch of problems here: there is no easy way to click inside the monaco editor and get its content, as when it gets rendered there's a lot of under the hood magic to make everything work as you see, the DOM is pretty complex and there is no easy way to access the full solution string. Copy and paste is not supported in cypress as of this moment, I've seen many people asking for that in issues but no solution yet.

So, I was thinking of injecting the lesson solution somewhere in the html in a `data-attribute` and get it from there, but we would still need to paste that in the monaco editor to submit the solution, so this doesn't solve that either. 

Have you got any suggestion @olizilla @terichadbourne?

terichadbourne:
Is Cypress only able to access DOM elements like a user would, or is it also able to access underlying code? It feels like what we need to do is run the `run` function passing in `solution` instead of `code`. If it can only access the DOM, could we make a new invisible button that the user never sees, but Cypress can, that initiates a new function to do that? (And also make this invisible to people using accessibility accommodations?) Better than that would be if Cypress can access the code itself so it can just initiate that function without interacting with the DOM. 

Alternatively, is there some way to tell when it's Cypress and not a human clicking a button? If so, we could insert a conditional in the existing `run` method that replaces `code` with `solution` if it's Cypress clicking the normal user-facing button.

fsdiogo:
> Is Cypress only able to access DOM elements like a user would, or is it also able to access underlying code?

As this is end-to-end testing I'd say Cypress was only supposed to access the DOM like a user would, but I'm not sure.

> If it can only access the DOM, could we make a new invisible button that the user never sees, but Cypress can, that initiates a new function to do that?

Probably, although I'm not a fan of that solution.

> Alternatively, is there some way to tell when it's Cypress and not a human clicking a button? If so, we could insert a conditional in the existing run method that replaces code with solution

That's a good idea, thank you 🎉 There is a way to [detect if the app is running under Cypress](https://docs.cypress.io/faq/questions/using-cypress-faq.html#Is-there-any-way-to-detect-if-my-app-is-running-under-Cypress), so I'll inject the solution instead of the code when calling the `run` function as you said!

@olizilla do you think the solution above is OK or is it an anti-pattern or something to avoid?

terichadbourne:
This is working for me with @fsdiogo's edit. 🎉 

@olizilla Am I remembering correctly that this is built to handle normal lessons but can't yet handle file upload lessons? Do we need to teach it to upload a file that's stored within the website itself?

fsdiogo:
@terichadbourne this is not ready for file upload lessons and Cypress doesn't support file uploads, although it seems there are some workarounds: https://github.com/cypress-io/cypress/issues/170

fsdiogo:
After a chat with @olizilla, we came to the conclusion that the best way to do this was to follow @terichadbourne's suggestion:

> could we make a new invisible button that the user never sees, but Cypress can, that initiates a new function to do that?

So, right now the flow of the test is the following (simplified version):

1. Enter lesson
2. Click `View Solution`
3. Click `Replace With Solution` (this is a special button only used by Cypress to replace the code with the solution)
4. Submit the code

terichadbourne:
@fsdiogo I've successfully run the Cypress test and also checked that as a keyboard tab user and a screen reader user (and myself) I can't spot the new secret button. I'm good with this if you are!

fsdiogo:
Cool, I'll merge this then 🚢

Issue #211, opened 2019-04-26
Bug: No intuitive way to go back once clicking "View Solution" button


jimcal:
**Describe the bug**
Once the user click `View Solution`, there is not an intuitive way to go back the solution input by the user.

**Does this bug apply to a specific lesson or tutorial, or does it appear to be sitewide?**
Currently, this applies to all the scenarios that has the button `View Solution`

**Expected behavior**
I understood that this feature is a much welcome feature, and was just merged two days ago. However, it would be nice to compare solution to the original answer by the user.

**To reproduce**

- Go to any lesson.
- Click on 'View Solution'
- Other than reset code (which would wipe out the user's input) there is no other way to go back to the state contains the solution user previously enter.

jimcal:
_Fully expect someone said this is a feature not a bug._

fsdiogo:
Hi @jimcal, thanks for pointing this out.

I definitely can see the benefits of having the ability to compare the user's code with the solution.

A way to solve this would be to have the solution appear in a split window, like you can do in any good code editor, but I have no idea if that's possible in monaco editor.

I'll take a look and see what I can do.

fsdiogo:
Potential solution in https://github.com/ProtoSchool/protoschool.github.io/pull/215, check it out.

jimcal:
@fsdiogo excuse me for the late response, but this looks great! Look forward to this getting merged!

fsdiogo:
@jimcal it has been merged! Thanks for the feedback!

PR #215, opened 2019-04-29
fix: preserve user code when viewing solution

fsdiogo:
With the lib we're using to render the monaco editor we can't open a new tab nor have a [diff editor](https://microsoft.github.io/monaco-editor/playground.html#creating-the-diffeditor-multi-line-example) (this would be the nicer solution IMO).

I've been looking for other vue monaco lib but haven't found any that supports that.

So, another approach I thought about was to toggle another editor (read only) with the solution:

sol

## See it in action

s

There are two problems that can arise: 

- When there's a lot of code in the editor, you'll have to scroll down to view the solution (to overcome this we can automatically scroll to the solution code when pressing the `View Solution` button);
- It may not be clear to the user what gets submitted when you press the `Submit` button;

<hr />

Closes https://github.com/ProtoSchool/protoschool.github.io/issues/211.

### DIFF
--- a/src/components/Lesson.vue
+++ b/src/components/Lesson.vue
@@ -83,10 +83,13 @@
           </div>
         </div>
         <div class="mb2">
-          <span v-if="solution" @click="viewSolution" class="ml1 fl textLink">View Solution</span>
+          <div v-if="solution">
+            <span v-if="viewSolution" @click="toggleSolution" class="ml1 fl textLink">Hide Solution</span>
+            <span v-else @click="toggleSolution" class="ml1 fl textLink">View Solution</span>
+          </div>
           <span v-if="cachedCode" @click="resetCode" class="mr1 fr textLink">Reset Code</span>
         </div>
-        <div class="h-100 flex-auto bg-white">
+        <div class="h-100 flex-auto">
           <MonacoEditor
             class="editor"
             srcPath="."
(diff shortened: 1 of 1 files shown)


terichadbourne:
@fsdiogo Just confirming that the way this now works, we are never overwriting the content of the code editor automatically, so a user would view the solution visually and then make updates to the code editor on their own, right? Personally I'm happy to force people to do this for a more effective learning experience, but I think it will mess up @olizilla's current plan for automatic testing where the robot clicks "View solution" and then "Submit." Maybe we can teach the robot how to copy and paste? :)

fsdiogo:
> Just confirming that the way this now works, we are never overwriting the content of the code editor automatically, so a user would view the solution visually and then make updates to the code editor on their own, right?

Right on.

> Personally I'm happy to force people to do this for a more effective learning experience

Agree, the other way around you could just mindlessly click `View Solution` and `Submit` without even looking at code, this way you're forced to look at it a bit more.

> but I think it will mess up @olizilla's current plan for automatic testing where the robot clicks "View solution" and then "Submit." Maybe we can teach the robot how to copy and paste? :)

Yes, it will mess with that 🤔 as I'm the one breaking it, I'm up for figuring out a way to solve it 🙃

fsdiogo:
Did some small tweaks to the UI as per the feedback of the Design Weekly Call:

- Positioned `View Solution` (added a chevron to mark that it's expandable) below the user code, so when you click it the solution appears below the button, so the feeling of action-reaction is more obvious
- Aligned `Reset Code` to the left to make code actions coherent and removed the text underline (that seems like a link that will open a new page)

terichadbourne:
@fsdiogo Looking good in terms of placement! I think the arrow behavior pattern that @dzesika was describing on that call, and which I'm familiar with from other sites, is actually a solid triangle that points right when the section is closed/hidden and down when it's open/revealed, like this: 

image

Would that be an easy enough fix to make the UI look more familiar?

fsdiogo:
@terichadbourne here it is:

chev

PR #218, opened 2019-05-01
Explain how to clear localStorage

terichadbourne:


terichadbourne:
@fsdiogo I've added a description of how to reset your localStorage so you can pretend to be a user with no history when testing.  Could you please read through this and merge if this explanation makes sense to you?

PR #223, opened 2019-05-10
feat: add markdown to outputs

fsdiogo:
Add markdown support for both the success and fail messages, so you can do whatever:

| old | crazy stuff you can do now |
| --- | --- |
| a | c |

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ if (result === 'foo') {
   return { success: 'Happy Message!' }
 } else if (result === 'bar') {
   return {
-    success: 'You did it!',
+    success: 'You did it!', // A string, which can optionally include markdown formatting
     logDesc: 'Check out the current value of `CID`:', // A description of the data you're displaying (a string, which can optionally include markdown formatting)
     log: ipfsFiles // The data you want the user to see (a JSON object or a string)
   }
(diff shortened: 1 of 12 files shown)


terichadbourne:
@fsdiogo Thanks for adding this feature!

There's not enough difference between these two shades of red for me to quickly tell one part is different from another: 
image

I tried flipping it to a success message and I can't see the contrast in the greens well enough either: 
image

Is there an easy way for you to change those colors so the formatting of code is more visible?

terichadbourne:
I also noticed that you changed some of the cypress code. Do you need me to run those tests as well?

fsdiogo:
> Is there an easy way for you to change those colors so the formatting of code is more visible?

Yeah I can easily tweak the colors. Do you have any suggestion?

> I also noticed that you changed some of the cypress code. Do you need me to run those tests as well?

No need, I just removed unused data-attributes.

terichadbourne:
Since the text is white, I think we should probably either switch the background to a much darker red/green for the code sections or perhaps switch it to a dark gray if that doesn't work?

fsdiogo:
Nice!

Pining to cluster is failing but as we can't trigger the build I'm going to merge this now and wait for cluster to be OK again to then merge https://github.com/ProtoSchool/protoschool.github.io/pull/224 and deploy these two PRs together.

Issue #99, opened 2018-12-07
"Useful Concepts" / Glossary UI improvements


terichadbourne:
In user testing, @meiqimichelle didn't love the placement of the "Useful Concepts" box, and felt that the information there (in this case definitions of CID and DAG) would be better suited to a glossary format where the user controls what information is displayed, or whether it's displayed at all. 

She shared an example of a glossary at  https://revenuedata.doi.gov/how-it-works/federal-reforms/ where the user picks what words they want defined in a sidebar, and can then also search for additional terms there. 

Although I like the glossary idea, I also think we should introduce some of these "useful concepts" in the lessons themselves. For example, in Basics Lesson 1 I recently changed the lesson content to include a very basic CID explanation that fit the context in which it was being used: 

>You can create a new node by passing a data object into the `ipfs.dag.put` method, which returns a Content Identifier (CID) for the newly created node.

>ipfs.dag.put({hello: 'world'})

>A CID is an address for a block of data in IPFS that is derived from its content. Every time someone puts the same `{hello: 'world'}` data into IPFS, they'll get back an identical CID to the one you got. If they put in `{hell0: 'w0rld'}` instead, the CID will be different.

Some of this will be alleviated by the new workshop @mikeal is building around decentralized data structures, which we may be able to link to in order to explain concepts like CID or DAG in more detail.

terichadbourne:
Noting also this issue for IPFS docs to add hover options for glossary: https://github.com/ipfs/ipfs-docs/issues/294

PR #225, opened 2019-05-14
chore: update useful concepts positioning

fsdiogo:
Reposition the `Useful Concepts` box as it was taking a lot of horizontal space from the lesson:

| old | new |
| --- | --- |
| old | new |

### DIFF
--- a/src/components/Lesson.vue
+++ b/src/components/Lesson.vue
@@ -9,11 +9,11 @@
             <span class="pl1"><img v-if="lessonPassed" src="../static/images/complete.svg" alt="complete" style="height: 1.2rem;" class="v-mid"/></span>
           </div>
           <h1>{{lessonTitle}}</h1>
+          <div v-if="concepts" class='fr-l measure-narrow-l ph3 mb2 ml3-l ba border-green' style="background: rgba(105, 196, 205, 10%)">
+            <h2 class="f5 fw2 green mt0 nb1 pt3">Useful concepts</h2>
+            <div class='f6 lh-copy' v-html="parsedConcepts"></div>
+          </div>
           <div class="lesson-text lh-copy" v-html="parsedText"></div>
-        </section>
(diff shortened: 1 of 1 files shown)


olizilla:
I deliberately pulled out the Useful Concepts box as aside to add a little visual interest, and as the lesson text discusses new concepts, it seemed better to pull the box up near the top, so they can see there is a glossary for those terms.

That said, it looks neat in the example... but perhaps it should just be explained in the body text of the lesson before the term is first used, rather than in a seperate box.

terichadbourne:
I agree with @olizilla that if it’s going to be used it should be an aside visible from the top, not moved to the bottom. However, we could try wrapping the main text so that we’re only losing the right side of the screen until the bottom of the concepts box and then the main text flows all the way to the right again. That would be my proposal for right now. 

See the proposal here for an in-line glossary feature (which we don’t have time to prioritize right now): #99

If we all agree that we don’t want the box used long term, we could remove the option from the instructions for building new tutorials and open an issue (to ignore until after camp) to go back and rewrite existing lessons that use it so they either define terms in line or link to external resources like the decentralized data structure tutorial.

fsdiogo:
What if we put it under the title?

terichadbourne:
@fsdiogo I like your suggestion of putting it at the top, but only for mobile with a very narrow screen. When the space is available, my own preference would be to leave it as an aside at the top right (visible from the top but without any suggestion that it's the main content) and fix the wrapping so that the main content can take up the full width of the screen once it's under the aside (old-school float style). I also think the top of the useful concepts box should be under the lesson title, not aligned with it. Here's a shoddily copy-and-pasted mockup of what I'd envision for when the real estate allows: 
image

fsdiogo:
I like your suggestion @terichadbourne 👍

My last commit addresses everything:

| Desktop | Mobile |
| ------- | ------ |
| desktop | mobile |

What do you think?

terichadbourne:
Looks great @fsdiogo. Thank you!

Issue #129, opened 2019-01-14
Feature: "Next steps" closing page for all tutorials.


mikeal:
At the end of a lesson we link back to the tutorials.

Instead, we should create a page full of potential next steps someone can take after learning what they've just learned. It could include:

* Documentation and blog posts about creating things with what they've learned.
* Entry points into relevant communities.
* In more advanced tutorials it could even link into places where people are hiring developers in this particular area.
* And of course, it should link in to places in the ProtoSchool project where they can contribute as well.

NukeManDan:
Also consider:
- example links to projects/apps that have used the concept learned for a real use case (eg. PeerPad w/ CRDTs / others in IPFS Awesome:  https://ipfs.io/ipfs/QmUQs8yuYbQ7VbMpRoL6RFz4dMuK4RqCCmPmJc5fK7P4Yc/)

terichadbourne:
@mkikeal Do we want this page accessible all the time from the main menu? So if a tutorial now has 3 lessons, it will instead have 4 and Lesson 4 will be "Next Steps," and it would be visible when you are viewing this type of listing? 

![image](https://user-images.githubusercontent.com/19171465/51275859-1ee0f880-19a1-11e9-8c6a-ca00cb3f6de2.png)

The other option would be a top secret page that's not surfaced anywhere but linked to from the last publicly visible lesson as a sort of reward for finishing (in which case we'd have to decide whether you only have to have finished the last lesson in the tutorial or all of them, and build some new code if the latter. 

I'm leaning toward making it accessible to navigate to in case anyone wants to jump right to recommendations for follow-up materials/projects. 

Thoughts?

mikeal:
Since it isn't something you can "complete" I'm leaning towards saying no, the last page should be hidden from this particular view.

terichadbourne:
My concern with making a next steps page appear only after you've completed a full tutorial, rather than being a visible part of the curriculum, is that users won't be able to navigate back to it. 

Per a discussion with @mikeal, we'll postpone addressing this issue until after making a decision on Vuepress (#141), which we'll be exploring the week of 2/11-2/15.

terichadbourne:
We'll be addressing this one in the near future, with @fsdiogo taking the first stab at it. Some of the goals we discussed today: 

- The "Next Steps" or "Resources" page will be visible from the main landing page for the tutorial, as the last lesson in the list (you will not have to complete the lessons to access it)
[...]

PR #226, opened 2019-05-16
feat: add resources lesson

fsdiogo:
<hr>

TODO:
- [x] Delete the dummy `Resources` in `Basics` tutorial
- [ ] Delete the dummy `Resources` in `Data Structures` tutorial

Closes https://github.com/ProtoSchool/protoschool.github.io/issues/129.


fsdiogo:
@terichadbourne when you have the time, please outline the format of the resources so I can change this PR!

fsdiogo:
@terichadbourne as we talked about in yesterday's call, the format of the `Resources` lesson is now strict.

A user can only edit the initial description, the links in the bullets and the final description.

The way to do it is through the `tutorials.json` file:

https://github.com/ProtoSchool/protoschool.github.io/blob/18f59cb7ded655bb9a2542a2a9d4c7d76a1c6568/src/static/tutorials.json#L14-L22

That gets fed to the page without the user having to copy or create files:

a

terichadbourne:
@fsdiogo This is looking great! Some questions/suggestions below...

**Formatting of lists**

Based on the discussions I had with people at IPFS Camp, it sounds like the following format for the list items on the resources page would be most helpful: 

[**Title of a Linked Resource in Bold**](some-linked-resource.com) {tag}
Optional description of resource 

[**Title of another Linked Resource in Bold**](some-linked-resource.com) {tag}
Optional description of resource 

This could either be bullet-free or have a bullet on the first line (but only if we can be sure the description gets the same indentation). 

The data the author would provide for each of these resources would be: 
- title of resource
- link to resource
- description of resource (optional) 
- type of resource (video, ProtoSchool, docs, article, etc.) - this would appear as a little tag - maybe white on a navy background so that people who love to learn by video, for example, could pick the stuff that interests them. We won't force people to pick from a specific list of tags now, just watch what they do add more structure later if we decide we need to

I attempted to make this change but it looks like you're creating markdown strings so I wasn't able to do the traditional v-if stuff in Vue. 

Right now there are commas appearing between the bullets that shouldn't be there, but I can't figure out how they're appearing so I can't remove them. 🤷‍♀️

We should probably call the list "resources" or "items" instead of "bullets" so that it's not related to our display choices.

*Consistent naming* 
I think "Resources" does make more sense here than "Next Steps," so I made one edit to that effect. 

**Tests**
Is this `hasResources` in `cypress/integration/tutorials.spec.js` something the author will have to update in the test suite or something that will work itself out automatically? 
```  viewSolutionsAndSubmitAll({ tutorialName: 'basics', lessonCount: 3, hasResources: true })```

**Instructions & mandatoriness**
Am I correct that the instructions for adding resources in the README.md file are from a previous iteration of this work? Mind updating them to the new method? 

My intention is for these pages to be mandatory moving forward. I'm flexible on whether we look for it manually when someone submits a PR or whether the system ensures it exists (once I've added content for the existing tutorials), but I don't think we should treat this step as optional in the instructions.

**Component structure in project**
Is there a reason that `landing.vue` and `resources.vue` aren't in the `components` folder? It took me a long time to find the resources component because that's where I was looking, and I was thinking of the `tutorials` folder as only containing the folders for each specific tutorial, created by the author, as opposed to the underlying building blocks. Thoughts?

fsdiogo:
> I attempted to make this change but it looks like you're creating markdown strings so I wasn't able to do the traditional v-if stuff in Vue.

Yeah, I changed this so we can now write in the regular Vue templating language.

> Right now there are commas appearing between the bullets that shouldn't be there, but I can't figure out how they're appearing so I can't remove them. 🤷‍♀️

I had this issue because I was mapping the array and returning strings, so the commas hanged on. This is now fixed due to my previous comment.

This is what we have right now:



> Tests
Is this hasResources in cypress/integration/tutorials.spec.js something the author will have to update in the test suite or something that will work itself out automatically?

Nope, the user doesn't have to change anything. This change is exactly to check if a lesson as a resource and to test it accordingly.

> Instructions & mandatoriness
Am I correct that the instructions for adding resources in the README.md file are from a previous iteration of this work? Mind updating them to the new method?

Probably yeah, I'll update the README.

> My intention is for these pages to be mandatory moving forward. I'm flexible on whether we look for it manually when someone submits a PR or whether the system ensures it exists (once I've added content for the existing tutorials), but I don't think we should treat this step as optional in the instructions.

We should define this now then, because if the Resources will be mandatory there are some check I'm doing that won't be needed. 

> Is there a reason that landing.vue and resources.vue aren't in the components folder? 

They are not shared components, so that's why I excluded them from that folder.

terichadbourne:
Looking good @fsdiogo! A couple of tiny CSS requests for that list of resources. 
- It looks like the tag is sitting lower than than the title - could those be vertically aligned? 
- Could we make less blank space between the title and its corresponding description so they look like more of a unit? (The space there should be less than the space between that description and the following title.) 
Thanks!

fsdiogo:
Done @terichadbourne!



Sometimes it may feel unbalanced if the middle item has no description, although the spacing is correct:

terichadbourne:
I like that reformatting, thanks!

fsdiogo:
> Is there a reason that landing.vue and resources.vue aren't in the components folder?

Thought a bit more about this and you are right. Changed the name to `ResourcesLesson` so it doesn't clash with the `Resources` component, but added theses two to the components folder.

Issue #177, opened 2019-03-27
Feature: Support multiple choice questions


terichadbourne:
Use this issue template to share a suggestion for a new feature or enhancement to be applied across the ProtoSchool site. (If you have feedback on a specific lesson, please use the Lesson Feedback issue template instead.)

**Is your feature request related to a problem? Please describe.**

@alanshaw has a new tutorial in mind for an upcoming event that would utilize multiple-choice questions instead of a code exercise.

**Describe the solution you'd like**
A new component would replace the exercise component and a builder of a lesson would have the option to build a multiple choice question with validation instead of a text-only lesson or lesson with code exercise or file upload + code exercise. 

The builder of the tutorial would need to establish: 
-  the question 
- some strings as the answers the user could select from
-  something indicating which is the correct answer
- error messages to display for each of the wrong answers (explaining why they're wrong) 
- success message to display for the correct answer (could help reinforce why it's right)

**Describe alternatives you've considered**

**Additional context**

Questions: 

Is it reasonable to force the builder to only work with radio buttons (single answer) as opposed to checkboxes (multiple answers) to simplify development? 

Should we define a set number of answers allowed (4?) or let the builder decide? 

Should the data go in as an array of objects to avoid a ton of strings? Something like this as options you'd put into the Vue file created for the lesson? 
{
question: "string question?",
answerOptions: [
{
answer: "string answer that's wrong",
correct: "false",
message: "friendly error message with hint to make them try again"
},
{
answer: "string answer that's right",
correct: "true",
message: "that's right because of this!"
},
{
answer: "string answer that's wrong",
correct: "false",
message: "friendly error message with hint to make them try again"
},
{
answer: "string answer that's wrong",
correct: "false",
message: "friendly error message with hint to make them try again"
}
]
}

Presumably this way we could loop through that data and print the options with how ever many answer letters are needed (A, B, C, D, etc.) and then use the "correct" field for validation purposes. 

The Lesson.vue component would get a v-if that only shows based on a new `isMultipleChoice` property that would be passed in (and replaces the exercise box). 

Would this meet your needs @alanshaw? 

How does this architecture strike you @olizilla @mikeal?

alanshaw:
Yes this sounds great!

I don't think we should limit the number of choices.

I was thinking about an optional `maxSelections` prop and have the UI determine the appropriate controls to render....but I think initially single answer questions (radios) would be fine.

How about renaming `answerOptions` => `choices`?

[...]

PR #240, opened 2019-06-14
feat: add multiple choice lesson

terichadbourne:
@fsdiogo @olizilla I've just started working on adding a multiple choice lesson type (see issue #177) but would love some on the way this fits into the project before I keep going.

Right now I'm using two new components: 
- Multiple-Choice-Lesson.vue (extends Lesson, basically just sets an "isMultipleChoiceLesson" property)
- Quiz.vue (takes in `question` and `choices` - passed from lessonnumber.vue to MultipleChoiceLesson to Lesson to Quiz - and returns HTML to make the radio buttons and figure out whether the user answered correctly and what feedback they need to see. 

I'm having trouble deciding whether to render the Quiz component in a similar place to where the exercise is normally rendered. At the moment it's done this way so that we can keep surrounding formatting, like the gray box, the status indicator, and the submit button. This means we wouldn't have to fix formatting in multiple places in the future. However, if we leave it this way I'd need to: 
- Figure out how to get the `feedback` and `answerStatus` from Quiz back up to Lesson and use them to create the feedback in the Lesson component instead of where it is in the Quiz component (otherwise I have to make the formatting much better in Quiz) 
- Change the "Submit" button so it's just a "Next" button, which maybe only appears after the answer is correct? You should get feedback immediately when you pick a radio button (true now) and there's no need for the validate or run functions. Not sure what's the best UX here. 
- Find the right way to set "lessonPassed" without using `run` or `validate`
- Change the labels with the status indicator at the top left of the gray box (just when it's a multiple choice lesson. For example, instead of "Try it!" -> "Keep working." -> "You did it!" it might be "Take the quiz." -> "Try again." -> "You passed the quiz!"
I guess the alternative to embedding this stuff conditionally in the lesson is that I'd copy a lot more code into the Quiz component, but then we'd need a way to keep formatting consistent through shared classes or something. What do you think? 

Some other questions: 
- Could we let authors use markdown in the `answer` and `choice.feedback` strings they provide?
- This is built so that the author needs to provide only one correct answer. We can specify that in the instructions, but is there a way to validate it as well? Maybe @olizilla's robot could learn to check that? Except you can't check it as a user, because behind the scenes it looks for the index number of the first correct answer and makes that the correct one. 

I'm still having fun being the one to work on this, but I'd love some tips to make sure I'm headed in the right direction! 

Note that I didn't surface my sample lesson to the listings, but you can find it at: http://localhost:3000/#/tests/01


fsdiogo:
Hey @terichadbourne, I'd say you're on the right track!

> Figure out how to get the feedback and answerStatus from Quiz back up to Lesson and use them to create the feedback in the Lesson component instead of where it is in the Quiz component (otherwise I have to make the formatting much better in Quiz)

For this you can use [`$emit`](https://vuejs.org/v2/guide/components.html#Emitting-a-Value-With-an-Event), I'll make a commit passing the data from the `Quiz` to `Lesson` component and you can carry on, as today I won't have time to work more on this. With that working, you'll just have to find a way to wire that data to the output so you can reuse the output formatting!

terichadbourne:
@fsdiogo thanks for setting up the `emit`! I discovered that it was transmitting the values from before the click instead of the values from after, so I changed @click to @change after reading up  and it seems to have fixed the problem: https://medium.com/@asselin/tech-note-using-vuejs-events-together-with-v-model-4334affce6e6 

I got the feedback switched so it comes through the normal place instead of inside the Quiz component (supporting markdown!), and it's setting lessonPassed when appropriate. However, I need to fix the status labels at top left about your status, make it know you haven't passed the lesson if you change your answer, and figure out how the button will look since we don't really need a submit option.

terichadbourne:
when done closes #177

fsdiogo:
**REMINDER:** before merging, delete the `tests` tutorial that was used to test this feature.

fsdiogo:
@terichadbourne we should update the `README` before merging this too, with instructions on how to use the multiple choice lesson.

terichadbourne:
@fsdiogo thanks for the reminder on the readme. can you please please review my edits there and confirm they make sense to you?

Issue #245, opened 2019-06-27
Bug: Unpredictable icon sizes on mobile 


terichadbourne:
When I hold my iPhone SE vertically, the icons on a tutorial homepage all appear at different sizes. I don’t have this problem when I hold it horizontally. 

![B22B7413-C06A-48A5-9124-9F314CFA886C](https://user-images.githubusercontent.com/19171465/60274020-9c694700-98f7-11e9-80f7-c66515af6f6c.png)

@fsdiogo do you have time to figure this one out? (Not urgent)

terichadbourne:
I see now that it’s not random. The more lines the text wraps onto, the smaller the icon becomes.

terichadbourne:
@fsdiogo Heads up that I'm reopening this issue because folks are still reporting varying icon sizes. @olizilla says "it looks like a flex issue, if the text is longer than space available, the space for the icon is flexing to accommodate it"

I'll add this to my list to take a stab at.

fsdiogo:
Yup, it needed `flex-shrink: 0` to keep the icon intact!

I've also truncated the lesson name to keep everything tidy:

<img width="384" alt="1" src="https://user-images.githubusercontent.com/33324750/60987072-c7608b80-a338-11e9-9a8f-052aadf7e925.png">

PR #246, opened 2019-06-27
fix: mobile icon sizes

fsdiogo:
se

Closes https://github.com/ProtoSchool/protoschool.github.io/issues/245.


fsdiogo:
I'm merging this as I've tested it in multiple browsers, on a mobile device and is working well.

PR #251, opened 2019-07-04
chore: update ipfs

fsdiogo:
- Update to the latest version of IPFS
- Update the hashes to base32 to pass validations
- Remove unneeded webpack configs

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -949,6 +949,346 @@
         }
       }
     },
+    "@hapi/accept": {
+      "version": "3.2.2",
+      "resolved": "https://registry.npmjs.org/@hapi/accept/-/accept-3.2.2.tgz",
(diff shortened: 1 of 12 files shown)


terichadbourne:
Thanks for this @fsdiogo! I've updated my npm packages and run locally and it seems to be in good order after testing just a few lessons. Can I assume that @olizilla's robot means I don't need to do any more specific checks? Or do I need to run through the file upload ones by hand?

This shouldn't change the 'passed' status of people who already submitted when it worked the other way, right?

terichadbourne:
@fsdiogo Have you already checked whether this version of IPFS includes the blob support in the browser? Is that resolved here or should we leave a separate issue open for it?

fsdiogo:
> Or do I need to run through the file upload ones by hand?

I ran it by hand just to safeguard and everything worked out as expected, so this one is good to go!

> Have you already checked whether this version of IPFS includes the blob support in the browser? Is that resolved here or should we leave a separate issue open for it?

I haven't tested it, but it should be done: https://github.com/ipfs/js-ipfs/issues/2024

terichadbourne:
Perfect, thanks for taking this one on!

PR #256, opened 2019-07-10
fix: mobile icons

fsdiogo:
Fixes https://github.com/ProtoSchool/protoschool.github.io/issues/245.

### DIFF
--- a/src/components/LessonLink.vue
+++ b/src/components/LessonLink.vue
@@ -3,12 +3,12 @@
     <div class="flex items-center">
       <div v-if="isResources" class="tc green ttu f6" style="min-width: 93px">Resources</div>
       <div v-else class="tc green ttu f6" style="min-width: 93px">Lesson {{index}}</div>
-      <div class="pr2 flex">
+      <div class="pr2" style="flexShrink: 0">
         <img v-if="lessonPassed('passed' + to)" src="../static/images/complete.svg" alt="complete" style="height: 1rem;"/>
         <img v-else-if="lessonCached('cached' + to)" src="../static/images/in-progress.svg" alt="in progress" style="height: 1rem;"/>
         <img v-else src="../static/images/not-started.svg" alt="not yet started" style="height: 0.9rem;"/>
       </div>
-      <div class="navy fw5 mw6">{{name}}</div>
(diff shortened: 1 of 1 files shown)


terichadbourne:
Thanks @fsdiogo. I actually can't test the icon sizes now that the titles are all one line, since the issue happened with wrapping. :) 

But on that note, I actually disagree with you about truncating. Imagine you use an iPhone 5 or SE and want to read one of our text-only lessons on the bus. There are only one or two words available per title if we truncate (see below) and as far as I know no way to see the full title without clicking through to load the page, since phones don't have a hover state to take advantage of. I for one would much prefer to have the line wraps so I can view the full titles to decide what content is worth my time to click through and load. Thoughts?

image

fsdiogo:
It's 100% true that you lose sight of a bunch of the text when truncating in mobile. Still, I prefer that than having a bunch of lines with different heights, it makes me itchy 🤷‍♂😛

That being said, I'll leave it to you to decide.

src/components/LessonLink.vue
```diff
@@ -3,12 +3,12 @@
...
       <div v-if="isResources" class="tc green ttu f6" style="min-width: 93px">Resources</div>
       <div v-else class="tc green ttu f6" style="min-width: 93px">Lesson {{index}}</div>
-      <div class="pr2 flex">
+      <div class="pr2" style="flexShrink: 0">
         <img v-if="lessonPassed('passed' + to)" src="../static/images/complete.svg" alt="complete" style="height: 1rem;"/>
         <img v-else-if="lessonCached('cached' + to)" src="../static/images/in-progress.svg" alt="in progress" style="height: 1rem;"/>
         <img v-else src="../static/images/not-started.svg" alt="not yet started" style="height: 0.9rem;"/>
       </div>
-      <div class="navy fw5 mw6">{{name}}</div>
+      <div class="navy fw5 mw6 truncate">{{name}}</div>
```
fsdiogo: Just remove the `truncate` here if you don't want to truncate the text.

terichadbourne:
I untruncated to favor usability over gorgeousness, but these identically-sized icons are looking lovely! 
image

Issue #179, opened 2019-03-27
Update hosting and URL strategy to allows analytics tools to distinguish between tutorials as separate pages 


terichadbourne:
Currently analytics tools are unable to distinguish between different views within our Vue SPA as though they're different pages. All page visits show as https://proto.school with no more specific URLs captured. Enabling more specific metrics collection would enable us to better understand user needs and trends to create a better experience for users in the future.

One suggestion from @mikeal is to use Zeit rather than GitHub Pages to compile our main website (in a way that would let chapter websites continue to publish via GitHub Pages for convenience,

terichadbourne:
@olizilla has suggested that [Countly](https://count.ly/) might be appropriate for this metric collection. (cc @mikeal)

terichadbourne:
Just adding a note here to remind myself that we should look at adding separate social metatags for each tutorial once we figure out how to get them treated as separate pages. I'm currently attempting to add this at a cross-site level in the meantime via #192.

terichadbourne:
Shared by @mikeal: we could use this to get the right card info (open graph metadata) on different pages https://github.com/chrisvfritz/prerender-spa-plugin

terichadbourne:
Have broken out the social sharing / metadata challenge into its own issue: #258. 

The original issue here to allow tracking by page view will be achieved when we merge #257.

terichadbourne:
closed via #257

PR #257, opened 2019-07-10
Add analytics w/ page views via Countly 

terichadbourne:
This PR adds tracking via Countly (hosted at countly.ipfs.io) while respecting "do not track" requests in the user agent. 

It taps into Vue's routing system to allow us to see the path for any resource that has its own path, such as tutorial landing pages and individual lessons: 
image

This PR also removes tracking via Google Analytics. (Will this cause any problems for you, @mikeal?)

This PR does not address event tracking, such as clicks on "reset code," "view solution," or "submit" buttons within lessons. We'll plan to address that in a second PR once this is in place. 

Note: If testing on localhost, you'll need to comment out line 70 in `index.html` (which restricts tracking to views of the live website) and uncomment line 65 (which tracks views on localhost as well).

Pair programmed by @olizilla & @terichadbourne

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -53,27 +53,39 @@
         }
         return dntStatus === 'Enabled'
       }
-      /* From https://github.com/nodejs/nodejs.org/blob/703bacbc0da63fa91f156468209758c66c01f30f/layouts/partials/footer.hbs */
+
+      //some default pre init
+      var Countly = Countly || {};
+      Countly.q = Countly.q || [];
+
       (function(){
+
+        // ************ FOR TESTING ONLY ************************
+        // Uncomment line below for testing, comment out for deployment
+        // if (!_dntEnabled()) {
+        // *******************************************************
+
+        // *********** FOR DEPLOYMENT *****************************
+        // Comment out line below for testing, uncomment for deployment
         if (!_dntEnabled() && location.hostname === 'proto.school') {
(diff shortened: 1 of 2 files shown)


terichadbourne:
Closes #179

terichadbourne:
@olizilla You mentioned something about potential cross-site issues with Safari. Should I run on localhost in Safari as the test for this and see if Countly catches my views? 

Was there something we need to do to check for when there's no Countly to reference? (Or maybe that's already dealt with by the order of the code?)

terichadbourne:
@olizilla I tried testing from Safari and no page views are registered. If I understood you correctly, this is because we're hosting the Countly instance on ipfs.io instead of proto.school, right? How much work would it be to create a separate instance on the other domain? I went to our existing Google Analytics and see that 16% of our visits over the past month (18% since launch) have come from Safari, so it's a reasonably sized chunk of data we'd be missing, including anyone who receives a link to ProtoSchool on Twitter or otherwise on an iPhone and clicks directly rather than making an effort to open their favorite browser. Should we plan on moving the Countly hosting? Perhaps we adjust this PR to keep GA enabled until that change is made?

terichadbourne:
@olizilla helped me figure out why my Safari visits weren't being tracked, and the setup is actually okay as is. After we merge we'll need to test again from Firefox to confirm the cross-site tracking is working (since there's an exception for localhost). If it's not, we'll need to host Countly on proto.school instead of ipfs.io.

Issue #255, opened 2019-07-10
Bug: Footer's not stuck to bottom of page


terichadbourne:
When my browser window is taller than the page content, the footer is halfway up my window and there's whitespace underneath it. 

![image](https://user-images.githubusercontent.com/19171465/60976043-f9eb9380-a2fa-11e9-90f7-60a4cfd8efa5.png)

This is most likely to actually be visible in our shortest lessons, like the resources ones we're adding. 

I'd like to see that whitespace appear above the footer instead of below it.

PR #259, opened 2019-07-11
fix: full height page

fsdiogo:
Make the page full height when there is no content.

Deleted some stuff for this screenshot just to prove that it works:
 
100vh

Fixes https://github.com/ProtoSchool/protoschool.github.io/issues/255.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -11,20 +11,35 @@ export default {
 </script>
 
 <style>
+
+html, body {
+  height: 100%;
+}
+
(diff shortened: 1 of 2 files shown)


fsdiogo:
Had to change the footer CSS because it was breaking in Safari 🤷‍♂

Tested in Chrome, Firefox and Safari ✅

terichadbourne:
Working for me as well. Thank you @fsdiogo!

PR #267, opened 2019-07-12
chore: refactor lesson component

fsdiogo:
👷🛠

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ node_js:
   - "10.15"
 addons:
   chrome: stable
+  apt:
+    packages:
+      - libgconf-2-4
 cache:
   # Caches $HOME/.npm when npm ci is default script command
   # Caches node_modules in all other cases
--- a/src/App.vue
+++ b/src/App.vue
@@ -54,6 +54,7 @@ pre code {
   padding: 0.6rem 0.8rem;
   display: block;
   font-size: 12px;
+  overflow: scroll;
 }
 
 .fill-current {
--- a/src/components/Breadcrumbs.vue
+++ b/src/components/Breadcrumbs.vue
@@ -1,0 +1,26 @@
+<template>
+  <div class="lh-solid v-mid f4 green">
+    <span class="b">{{workshopShortname}}</span> |
+    <span v-if="isResources">Resources</span>
+    <span v-else>Lesson {{lessonNumber}} of {{lessonsInWorkshop}}</span>
+    <img v-if="lessonPassed" src="../static/images/complete.svg" class="dib ml1 v-mid" alt="complete" />
+  </div>
+</template>
+
+<script>
+export default {
+  props: {
+    isResources: Boolean,
+    workshopShortname: String,
+    lessonNumber: Number,
+    lessonsInWorkshop: Number,
+    lessonPassed: Boolean
+  }
+}
+</script>
+
+<style scoped>
+img {
+  height: 1.2rem;
+}
+</style>
--- a/src/components/Button.vue
+++ b/src/components/Button.vue
@@ -18,22 +18,23 @@ export default {
 </script>
 
 <style scoped>
-  button {
-    opacity: 0.9;
-    min-width:120px;
-  }
-  button:hover {
-    opacity: 1;
-  }
+button {
+  opacity: 0.9;
+  min-width: 120px;
+}
 
-  button:active {
-    opacity: 0.95;
-    box-shadow: inset 0 0 8px rgba(0,0,0,0.3);
-  }
+button:hover {
+  opacity: 1;
+}
 
-  button:disabled {
-    opacity: 0.8;
-    box-shadow: initial;
-    cursor: default;
-  }
+button:active {
+  opacity: 0.95;
+  box-shadow: inset 0 0 8px rgba(0,0,0,0.3);
+}
+
+button:disabled {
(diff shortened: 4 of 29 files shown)


fsdiogo:
So, I've made some changes to the Validator (Submit/Next buttons) workflow:

- Previously, if you clicked `Reset Code` and then `Submit`, nothing visible would happen, and only after a code change you'd see the output. That is now fixed, the output will appear after you click `Submit`.
- When returning to a previously passed file upload lesson, if there are no files uploaded, the user now gets a message stating that in order to see the output he must upload files and resubmit (a submit button has been added to enable that). He can still click `Next` without changing or submitting anything to skip to the next lesson (this only applies to FileLessons). See it in action: 

> FileLesson

- There were some incongruences with the state when a lesson was previously passed but that code was modified to an incorrect one: as the lesson was marked as passed, even if the code was wrong, you would still see the `Next` button. Now, if the code is changed and wrong, the `Next` button will turn into a `Submit` one automatically, so the state of the lesson is always up to date.

@terichadbourne this hopefully fixes everything we talked about, can you please take a tour through the PR to check if this can be merged? 🙏

terichadbourne:
@fsdiogo 
For some of my already-passed-but-now-file-less-lessons, it's doing what you intend: 

image

and for others it's putting both buttons but no special message above them: 

image

Notice that on that second one it doesn't have the green checkmark above the code, even though it did when I selected it from the table of contents. Ooooh, so I must have deleted the code, so why is there still a Next button and not just a Submit button?

I was also able to get a Next button even though I submitted incorrect code (with files included) in a previously-passed file lesson. Is it failing to wipe the result for some reason?

image

There's one unique lesson, blog #2, where all they're supposed to do is delete a comment and hit submit (I think the reason I had them delete the comment was that previously we needed a text edit to the code box in order to make the evaluation possible. Now when you do that, or just hit submit without deleting the comment, which should produce the same result code-wise, it's stuck on a Submit button instead of a Next button. There's no positive or negative feedback, just the log results and the Submit button.

image

The non-file lessons appear to now use the wrong header for the Update Code section. They should only say "Update Code", not "Step 2: Update Code." That phrase should be reserved for the file lessons, where a Step 1 also appears. 

image

Did you do anything that would affect our success.txt file? This lesson is not validating when I use the correct code, although it does give me error messages when I use incorrect code: 

image

Note, when the loader gets stuck, my instinct to try to stop it is to hit Reset Code. Could we make it so that clicking that stops the run function if it's in progress?

fsdiogo:
> Notice that on that second one it doesn't have the green checkmark above the code, even though it did when I selected it from the table of contents. Ooooh, so I must have deleted the code, so why is there still a Next button and not just a Submit button?

When you delete the code or just a character, the lesson gets unmarked as passed and only the `Submit` button appears. That's what is happening on my machine 🤔

> I was also able to get a Next button even though I submitted incorrect code (with files included) in a previously-passed file lesson. Is it failing to wipe the result for some reason?

From what I've been testing it's working as it should. Not sure what's happening.

> There's one unique lesson, blog #2, where all they're supposed to do is delete a comment and hit submit (I think the reason I had them delete the comment was that previously we needed a text edit to the code box in order to make the evaluation possible. Now when you do that, or just hit submit without deleting the comment, which should produce the same result code-wise, it's stuck on a Submit button instead of a Next button. There's no positive or negative feedback, just the log results and the Submit button.

That's happening on proto.school too 😓 I'll take a look.

> The non-file lessons appear to now use the wrong header for the Update Code section. They should only say "Update Code", not "Step 2: Update Code." That phrase should be reserved for the file lessons, where a Step 1 also appears.

Nice catch, I've fixed it!

blg

> Did you do anything that would affect our success.txt file? This lesson is not validating when I use the correct code, although it does give me error messages when I use incorrect code:

The MFS lesson 10? Nope, that logic is untouched. I just tried it and everything is working, both the errors and the success 😩

| error | success |
| ----- | ------- |
| err | suc |

> Note, when the loader gets stuck, my instinct to try to stop it is to hit Reset Code. Could we make it so that clicking that stops the run function if it's in progress?

The loader isn't supposed to get stuck with the changes made in this PR. I don't think we can do that: I can disable the loading when clicking `Reset` but the code will still be validating and when finished, it will change the page accordingly. So it would be pretty weird you canceling and then things changing again a bit after. I'd prefer for it to stay this way.

fsdiogo:
> There's one unique lesson, blog #2, where all they're supposed to do is delete a comment and hit submit (I think the reason I had them delete the comment was that previously we needed a text edit to the code box in order to make the evaluation possible. Now when you do that, or just hit submit without deleting the comment, which should produce the same result code-wise, it's stuck on a Submit button instead of a Next button. There's no positive or negative feedback, just the log results and the Submit button.

Ok, this was funny, I've also tripped on it 😄

Take a look at the description of the exercise:

ex

So, deleting that comment was just so we can see the output, it's not the solution to the exercise 😂

I've deleted that part because we can now just `Submit` the exercise as is to see the output.

fsdiogo:
> I was also able to get a Next button even though I submitted incorrect code (with files included) in a previously-passed file lesson. Is it failing to wipe the result for some reason?

I've figured out how you managed to do this: you returned to a passed lesson and clicked `Submit` without uploading files again. It should always clear the passed when you submit wrong code, not sure why it wasn't doing that sometimes.

PR #285, opened 2019-08-14
feat: add Countly app keys

fsdiogo:
We created a new app in Countly to track events when in development mode. We need this so we don't pollute the main app in Countly with dummy and testing events.

This PR chooses which key to use accordingly.

**PS:** I wanted to do this using env variables, but as Countly seems to have to be inited in the html file, I can't use env vars there. 

68747470733a2f2f636f756e742e6c792f6769746875622f636f756e746c792d7765622d73646b2e706e67

Ref: https://github.com/Countly/countly-sdk-web#how-to-use-countly-web-sdk


terichadbourne:
This is working for me, but we need to now remove these 3 commits before merging: 

image

Issue #269, opened 2019-07-16
Migrate Countly to proto.school domain


terichadbourne:
As of #257, we're now using countly.ipfs.io for our analytics. Although this is successfully capturing views from Chrome and Firefox, it only works with Safari when the user unchecks the apparently default "prevent cross-site tracking" option in their preferences. This appears to be the default on both desktop and mobile, so we're likely to miss most users of iPhones (if they just click a link from Twitter, for example, instead of opening up their preferred browser and pasting in a link). It sounds like Firefox may also be working on implementing a similar default.

Presumably moving our Countly service to the proto.school domain would prevent this gap in our analytics.

terichadbourne:
@mikeal I assume @olizilla and I will need your help with this one since you still own the proto.school domain. If I'm summarizing my chat with Oli correctly, the hope is that we would be able to do the following to appease the gods of cross-site tracking: 
- leave proto.school pointed to our existing gh-pages server for the ProtoSchool website content
- set up another domain/alias in DNS (countly.proto.school) pointing it to the same IP address as our current Countly deployment 

@olizilla Can you do whatever translating is necessary to make that request ^ accurate? 😂

mikeal:
What’s the IP address of the the countly deployment?

mikeal:
I found code in the tree point at `countly.ipfs.io` so I created a CNAME from `countly.proto.school` to `countly.ipfs.io`. It’ll take a little while to propagate but should work soon.

olizilla:
That is correct! I haven't spoken to countly yet. Hopefully it's as simple as pointing other domains at it, but I need to check.

terichadbourne:
Thank you both! I don't know if this is evidence that it will do exactly what we need, but I can now log into the same dashboard at countly.proto.school that I saw at countly.ipfs.io. @olizilla I'll wait until you hear back from Countly and then I presume we'll need to change a few details under the hood to make the requests go to the right place.

olizilla:
Good news! The change we need is this line https://github.com/ProtoSchool/protoschool.github.io/blob/26906e08712c9aa5f0476bc1be84781cf65561de/public/index.html#L75 to point to countly.proto.school

Countly will have to set up a new SSL certificate for that domain for it to work on https://countly.ipfs.io, but I suspect they use let's encrypt, so it might not be a big deal.

terichadbourne:
[...]

PR #288, opened 2019-08-16
fix: countly sdk src

fsdiogo:


terichadbourne:
actually fixed issue #269 after PR failed to deploy #276

Issue #289, opened 2019-08-16
Reference tutorials by IDs, not shortnames, when caching to localStorage


terichadbourne:
Fix things under the hood so that if we change the name or shortname of a tutorial, the way we cache code in localStorage won't be affected and if a user previously passed it under a different name it will still be passed.

terichadbourne:
This will also require a retroactive fix so that folks who have the currently existing tutorials cached have those caches relabelled in localStorage.

terichadbourne:
As part of this work let's add a clear section to the instructions on the various places things need to be updated if renaming a tutorial. 

We should also add a checklist for maintainers about what to check for when a new tutorial is submitted (for example, adding to the list of tutorials to be checked by Cypress if relevant). That list could potentially live here but be linked to from the `admin` repo.

fsdiogo:
I'm taking a look at this one, and I've found some caveats:

- the tutorial shortname is always constructed based on the path (the tutorial URL), so we don't have it stored anywhere, and even if we did it wouldn't help us because we use the current URL to add it to the navbar and more;
- the keys we're saving in localStorage are based on the tutorial shortname, i.e, the URL, so **the paths are what dictate the shortname and the localStorage keys**;

So, I'm not sure how to create a unique ID and decouple that from the tutorial path, as even in the Lesson component all the localStorage logic is using the current path and nothing more.

Is it OK to ask a tutorial creator to create an URL knowing that the shortname will be derived from it?

PR #292, opened 2019-08-21
feat: rename tutorials

fsdiogo:
TODO:
- [x] Revert the tutorial `002` url from `xxx` back to `basics`
	- [x] main.js
	- [x] tutorials.json
- [x] Comment the line 104 on `main.js`
- [x] Write instructions on the README on how to rename a lesson

Fixes https://github.com/ProtoSchool/protoschool.github.io/issues/289.


terichadbourne:
@fsdiogo The navbar is broken for me, for example clicking "tutorials" from the nav did tried to go to this: http://localhost:3000/#//tutorials

Cache appears to be updating properly for me when I migrate a URL. 🎉

Have you already looked at whether there's a reasonable way to create redirects in case someone shares an old link?

terichadbourne:
Speaking of redirects, this is interesting....
404 if you try to load something that doesn't exist with no hash in the URL 
image

versus blank page if you try to load the same non-existent thing with a hash 
image

Although ideally we'd do even better and more specific redirects for tutorials whose names have changed, it would be great to also have a backup that defaults any non-existent things to our homepage. I may not have read the Vue documentation correctly the first time around when I did this: 
```js
// 404
  { path: '*', name: '404' }
```

There's redirect documentation here if you have time to take a look: https://router.vuejs.org/guide/essentials/redirect-and-alias.html#redirect

fsdiogo:
> @fsdiogo The navbar is broken for me, for example clicking "tutorials" from the nav did tried to go to this: http://localhost:3000/#//tutorials

Hmm, it's working here. Can you tell me the steps you took?

> Cache appears to be updating properly for me when I migrate a URL. 🎉

Awesome 😎 

> Have you already looked at whether there's a reasonable way to create redirects in case someone shares an old link?

Haven't looked into it yet.

fsdiogo:
> Although ideally we'd do even better and more specific redirects for tutorials whose names have changed, it would be great to also have a backup that defaults any non-existent things to our homepage. I may not have read the Vue documentation correctly the first time around when I did this:

Redirecting a 404 to the home page is easy:

```js
{ path: '*', name: '404', redirect: '/' }
````

Although it's not the best experience having the user write a URL and be redirected to the homepage with no warning whatsoever. It would be best to create a simple 404 page.

fsdiogo:
Added a 404 page when we hit routes that don't exist:

fsdiogo:
Ooops, pushed the 404 page directly to `code` 😅 

I'll revert it and push it here.

terichadbourne:
@fsdiogo I see your note about making the 404 page but not a commit. Have you not yet pushed that? 

> @fsdiogo The navbar is broken for me, for example clicking "tutorials" from the nav did tried to go to this: http://localhost:3000/#//tutorials

> Hmm, it's working here. Can you tell me the steps you took?

1. Open website (localhost:3000/#/)
2. Click on any link in navbar 😂

Once you're inside a specific tutorial, these breadcrumb nav links work: 
image

But I can't find any page from which this main navbar works:
image
It always adds a second slash after the hash like so: 
image

fsdiogo:
Yikes, after restarting the server I'm getting the routes wrong! I'll check what's up.

PR #294, opened 2019-08-29
Missing await in the exercise 10 code (Mutable File System)

sericaia:
There was a missing `await`, which I believe was not on purpose.


terichadbourne:
Nice catch, @sericaia! That definitely looks like an oversight. Thanks so much for submitting the fix!

Issue #252, opened 2019-07-06
Lesson Feedback: Mutable file system - Lesson 4 (Add a file to MFS)


NukeManDan:
Have a question or suggestion regarding a specific ProtoSchool lesson? Please use this template to share it!

**URL of the lesson that's confusing:**
eg https://proto.school/#/mutable-file-system/04

**What's confusing about this lesson?**
Malformed solution leaves the "submit" button hanging.

```javascript
const run = async (files) => {
  for (let file of files) {
    await ipfs.files.write('/'+file, file, { create: true })  }
}

return run
```
Note the file name is malformed : '/'+file 

**What additional context could we provide to help you succeed?**

An "incorrect" prompt vs a loading loop.

Here is my console: 
![image](https://user-images.githubusercontent.com/35669742/60750438-9d3c5180-9f65-11e9-8b46-10d99db1b08d.png)

terichadbourne:
Thanks for documenting this @NukeManDan! 

_Apologies for what has turned out to be a very long braindump as I experiment..._ 

I initially read your comment as though the malformedness was about not putting spaces before and after the `+`, but this works: 

```js
const run = async (files) => {
  for (let file of files) {
    await ipfs.files.write('/'+file.name, file, { create: true })
  }
}

return run
```
so I can tell that the error is using just `file` (an object) instead of `file.name` (a string). I do get the same endless loading as you when doing it like this: 
```js
    await ipfs.files.write('/'+file, file, { create: true })
```
and in my browser (Chrome) the error is: 

![image](https://user-images.githubusercontent.com/19171465/64460054-74e8e480-d0c7-11e9-9078-02638c8d67a7.png)

It's getting stuck on trying to return `result.error` so it can print the error generated by the system.  Here's the relevant code: https://github.com/ProtoSchool/protoschool.github.io/blob/code/src/tutorials/0004/04.vue#L52

I tried logging just `result` in the console and got this: 

![image](https://user-images.githubusercontent.com/19171465/64460167-bd080700-d0c7-11e9-8d3d-75f39656c664.png)

So I assume `result` doesn't ever get passed from [the `run` function](https://github.com/ProtoSchool/protoschool.github.io/blob/code/src/components/Lesson.vue#L273) in the Lesson component to [the `validate` function](https://github.com/ProtoSchool/protoschool.github.io/blob/code/src/tutorials/0004/04.vue#L17) in this specific lesson's file. 

[...]

PR #301, opened 2019-09-20
refactor: return errors if errors are thrown

### LINKED ISSUE #337: Validation Error: Mutable File System | Lesson 7 (Create a Directory)
**URL of the lesson where you received the validation error:** https://proto.school/#/mutable-file-system/07

**The text of the validation error message you saw:** 
"Looks like you edited the ls code to list something other than the root directory. Please try again, editing only the section of code indicated."
image

**The code you submitted that caused this error message to appear:**
One line: `await ipfs.files.mkdir('/some/stuff')`
Full Code:
```js
  // This code adds your uploaded files to your root directory in IPFS
  await Promise.all(files.map(f => ipfs.files.write('/' + f.name, f, { create: true })))

  await ipfs.files.mkdir('/some/stuff')

  let rootDirectoryContents = await ipfs.files.ls('/', { long: true })
  return rootDirectoryContents
}

return run
```

Basically by omitting the `{ parents: true }` option.

achingbrain:
Makes the following changes:

1. Return errors from `_eval` instead of `{ error: Error }`
1. If a user's solution has a syntax error, return a special subclass of Error so we can show a `SyntaxError: ...` message to give them more of a clue what has gone wrong
1. Do not require custom solution validation to return something - if nothing is returned and the users' solution resulted in an error, show the error message, otherwise show a generic error message
1. Show the type of error in the output - e.g. `ReferenceError: ...`, `TypeError: ...`, our newly invented `SyntaxError`, etc, to help the user out

Also handles a few more types out output to the MFS.004 tutorial..

Refs: #252

Closes #337


terichadbourne:
Thanks @dominguesgm. I'll wait to do my own review until you've explored that.

achingbrain:
> When returning an undefined variable in lesson 9 of tutorial 4, it showed a success message (with the error text in the result)

Interesting.  I haven't see the success message with error text, but when I changed my solution to that tutorial to:

```javascript
/* global ipfs */

const run = () => {
  return undefined
}

return run
```

It dropped an error log into the console that it couldn't find a file but didn't show the user anything.  This was because the validation function tries to `ls` `/some/stuff` without handling the case when it doesn't exist.

I've updated the PR to handle validation functions throwing - it'll now show the user a generic error message asking them to reset their solution, as our validation functions shouldn't throw, they should return a useful message that helps the user work out what's wrong.

dominguesgm:
I think the issue comes from the fact that we sometimes override errors in order to handle them in the validation.

We sometimes need to handle IPFS errors in validation so we can give more meaningful feedback. However, if the error we get is a syntax error, we don't want to override the error message. I should've explained myself better in my previous comment but what I meant was that syntax errors were leading to a success message.

I'll be focusing on this in the next few days 👍

dominguesgm:
I think the issue comes from the fact that we sometimes override errors in order to handle them in the validation.

We sometimes need to handle IPFS errors in validation so we can give more meaningful feedback. However, if the error we get is a syntax error, we don't want to override the error message. I should've explained myself better in my previous comment but what I meant was that syntax errors were leading to a success message in that case.

Looking at the current approach, I would suggest the following:

 - Go through validation whether the result an error or not
    - If it isn't an error, validation should return something (success, failure)
    - If it is an error we want to improve the message, validation overrides it and returns the improved error
    - If it is an error we don't want to mess with (e.g. syntax), don't return anything
 - Test if the original result is an error and the validation result is undefined
    - If it is, do something generic with that error
 - If the validation result is not undefined, use its result

How would you feel about this approach? I like the idea of giving full control to the validation function but, by default, if it doesn't return anything, the `run` method will know what to do to cover for the lack of feedback.

I'll be focusing on this in the next few days 👍

dominguesgm:
I've ran into an issue when I was trying to tackle this problem. Namely, we currently do not have a way of knowing in the `run` method if the hypothetical error the `_eval` function throws will be overridden in the validation method.

I've come to two possible solutions which affect the way the user writes validation code:

- Either the user has to return something from the validation, success, failure or error (and this way we go back to the approach we had before this PR, with the addition of protecting against errors thrown inside the validation code). If he doesn't, we display a generic error about something being wrong in the validation

- Or the user is not required to return something, specifically in case of an error. For this, we need to have a way to inform the run method if the Error message has been overridden or not. If it has not been overridden, we ignore the output of the `validate` method and assign the original error message to `test`. Otherwise, we assign the new error message which comes in the output of the `validate` method.

A way to 'inform' the `run` method if the error message has been overwritten or not would be to have a new type of return value from the `validate` method. Instead of only:

```javascript
{
    success: message
}

or

{
    fail: message
}
```

we could have an extra

```javascript
{
    error: message
}
```

to signal the error message has changed. If after the `_eval` function returns an error, the output of the `validate` function is anything other than `{ error: message }`, we display the original error message. Of course, if the original result of `_eval` is not an error and `validate` returns undefined, we would still have a generic error message about something being wrong with the validation

I wanted to know your opinion, @terichadbourne and @achingbrain , on which approach you would feel is most user friendly when writing a lesson.

terichadbourne:
@dominguesgm When you say `user` in these two options, do you mean the learner (end user) has to have a return value in the function that they write or that the author of the tutorial who's creating the validation code needs to be super careful when writing their conditionals?

> Either the user has to return something from the validation: success, failure or error (and this way we go back to the approach we had before this PR, with the addition of protecting against errors thrown inside the validation code). If he doesn't, we display a generic error about something being wrong in the validation
> 
> Or the user is not required to return something, specifically in case of an error. For this, we need to have a way to inform the run method if the Error message has been overridden or not. If it has not been overridden, we ignore the output of the validate method and assign the original error message to test. Otherwise, we assign the new error message which comes in the output of the validate method.

I would theoretically be okay with forcing all lessons to be built in a way that requires the learner to return a value from their code (which isn't the case now, if we have ways to test the state of their node without anything returned), but I suspect I'm not answering your question.

dominguesgm:
@terichadbourne sorry, my bad for being confusing. I actually meant the tutorial developer and his experience while writing validation code. So it's a matter of the developer being forced to return something necessarily (and if he doesn't, throw a generic error), or let the developer not return anything (and if there is an error in the learner's code, throw it, if there isn't, throw a generic error).

dominguesgm:
This is ready to be merged, if anyone wants to give it a quick review first that would be great 👍

terichadbourne:
@dominguesgm I've proposed a few text edits (now committed) that I'd appreciate you checking for me. One of them is updating the way you check for the presence of existing IPFS-generated error messages to override, since I noticed you changing the format in specific lessons but didn't see that change in the instructions.

I assume you've found a quick way to force the validation error to happen for testing purposes. Can you please test that the link I put in does indeed render and go to the [new Validation Error issue template](https://github.com/ProtoSchool/protoschool.github.io/issues/new?assignees=&labels=lesson-feedback%2C+validation-error&template=validation-error.md&title=Validation+Error%3A+%5BLesson+Title+%28URL%29%5D) but with the lesson title and URL pre-filled? 

Can you please help me understand the difference between the circumstances that would lead to [this unknown error that mentions validation](https://github.com/ProtoSchool/protoschool.github.io/blob/a341ffe953d6b7a0aba9cfb771cbbf8b918f526d/src/components/Lesson.vue#L346-L351) and [this other unknown error that references validation and has the issue filing link](https://github.com/ProtoSchool/protoschool.github.io/blob/a341ffe953d6b7a0aba9cfb771cbbf8b918f526d/src/components/Lesson.vue#L360-L368)?

src/tutorials/0004/04.vue
```diff
@@ -37,19 +57,24 @@ const validate = async (result, ipfs) => {
...
-    return { fail: 'Uh oh. It looks like you created a directory instead of a file. Did you forget to include a filename in your path?' }
-  } else if (result && result.error.message === 'file does not exist') {
-    // Forgot the `{ create: true }` option
-    return { fail: "The file doesn't exist yet, so you need to create it. Did you forget an option?" }
   }
 
-  // Output the default error if we haven't caught any
-  return { error: result.error }
+  if (!ipfsFiles.length) {
+    return { fail: 'Uh oh. There was nothing in your MFS. Did you add the uploaded files to it?' }
```
terichadbourne:
```suggestion
    return { fail: 'Uh oh. There was nothing in your MFS node. Did you add the uploaded files to it?' }
```

src/tutorials/0004/04.vue
```diff
@@ -37,19 +57,24 @@ const validate = async (result, ipfs) => {
...
+  }
+
+  if (!itemsAreFiles) {
+    return { fail: "Looks like you didn't upload files. Did you upload a directory?" }
```
terichadbourne:
```suggestion
    return { fail: "Looks like you didn't upload any files. Did you upload a directory instead?" }
```

src/tutorials/0004/07.vue
```diff
@@ -52,7 +52,7 @@ const validate = async (result, ipfs) => {
 
   // Check whether array contains certain directory hash & certain directory name
   function contains (certainHash, certainName) {
-    if (!result.error) {
+    if (Array.isArray(result)) {
```
terichadbourne: Doesn't this need to be after the conditional for there being no result so that we don't run into an error about it being undefined?

src/tutorials/0004/07.vue
```diff
@@ -139,10 +142,6 @@ const validate = async (result, ipfs) => {
       logDesc: "Here's what was returned by `ls` in your root directory. Notice how directories have a type of `1` while files have a type of `0`.",
       log: logRoot
     }
-  } else if (result.error) {
-    return { error: result.error }
-  } else {
-    return { fail: 'Something we haven\'t anticipated is wrong. :(' }
```
terichadbourne: @dominguesgm Can you please confirm which error message now built into Lesson.vue will get surfaced now that this is gone? (This question applies across multiple lessons.) We have a number of different default unknown error messages across various lessons and I want to confirm we're putting the most appropriate one in the right place in the new shared code. Thanks!

terichadbourne:
@achingbrain I can't add you as a reviewer since you authored the PR, but I'd appreciate it if you could look through our edits tomorrow or Wednesday to ensure everything looks good to you with this revised approach before we merge.

src/tutorials/0004/07.vue
dominguesgm: This is just the function declaration, we are only using the `contains` function after checking if the result is not undefined.

src/tutorials/0004/07.vue
dominguesgm: In this specific lesson, with the logic in the previous 2 if blocks, this final else would never be reached. However, if there is missing validation, and the whole validation function is executed without a `return` statement being hit, we get the following message:

```javascript
} else if (test === undefined) {
  let validationErrorMessage = `You may have uncovered a bug in our validation code. Please help us improve this lesson by [opening an issue](${this.validationIssueUrl}) noting that you encountered a validation error and pasting in the code you submitted.`
  // Our validation did not return anything and the original result is also not an error.
  // This may be the result of a missing validation case + not returning anything by default
  test = {
    fail: validationErrorMessage
  }
}
```

This is our new equivalent to _'Something we haven't anticipated is wrong. :('_

terichadbourne:
@dominguesgm Just checking in to confirm you feel this one is good to go? 

I spoke with @achingbrain yesterday and he's planning to take a final look before we merge. Anything specific you want to make sure he pays close attention to?

dominguesgm:
Yup, I think it's good to go 👍

terichadbourne:
Just a quick reminder that this one is awaiting a final proof from @achingbrain to approve @dominguesgm's edits. Looking forward to getting this merged!

achingbrain:
LGTM 👍

src/tutorials/0004/07.vue
```diff
@@ -15,135 +15,7 @@ import text from './07.md'
...
-      success: 'Success! Check out your directory contents below.',
-      logDesc: "Here's what was returned by `ls` in your root directory. Notice how directories have a type of `1` while files have a type of `0`.",
-      log: logRoot
-    }
-  } else if (result.error) {
-    return { error: result.error }
-  } else {
-    return { fail: 'Something we haven\'t anticipated is wrong. :(' }
-  }
+  return null
```
terichadbourne: It looks like someone accidentally committed the version of `src/tutorials/0004/07.vue` that we changed for testing, replacing all of the validation code with `return null`. We could copy and paste the pre-this-PR version to replace this, but I'm not sure if there were other intentional changes made to this file before the accidental one. @zebateira @dominguesgm does one of you know the best way to address this on GitHub or locally? Is there a way to revert just one file's worth of changes one commit at a time until it goes back to having the last reasonable content? I'd love to be taught a good approach to this if either of you have any insight.

dominguesgm: Oops, my bad, great catch @terichadbourne 😅.

The best approach in this case (other than simply checking the past commit version and copy-paste it), is to checkout the previous version of the file with this command:

`git checkout HEAD~1 -- src/tutorials/0004/07.vue`

Btw, I also just remembered I had somewhat tackled issue #337 here. @zebateira do you want to give some feedback if you think the change made in mfs' lesson 7, line 81, is enough to fix the issue?

dominguesgm:
Btw, I also just remembered I had somewhat tackled issue #337 here. @zebateira do you want to give some feedback if you think the change made in mfs' lesson 7, line 81, is enough to fix the issue?

zebateira:
> Btw, I also just remembered I had somewhat tackled issue #337 here. @zebateira do you want to give some feedback if you think the change made in mfs' lesson 7, line 81, is enough to fix the issue?

@dominguesgm yes! it's perfect now 👌

PR #314, opened 2019-10-28
feat: protect code execution from hanging

dominguesgm:
If a user calls any IPFS API method with an incorrect `CID` (read non-existing), it will cause the execution to stall. For this reason, this PR covers the execution of the user code with a timeout.

This will abort the execution of the user code after 3 seconds and throw an error with the message:

`Your code took too long to execute. This may be a result of using a non-existing CID`

Though this is not the only case which will cause user code to hang (an infinite while loop would cause the same effect), I feel this message should provide the hint for a common error case (pasting the wrong `CID` into an IPFS API method call).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1316,11 +1316,6 @@
       "integrity": "sha512-shAmDyaQC4H92APFoIaVDHCx5bStIocgvbwQyxPRrbUY20V1EYTbSDchWbuwlMG3V17cprZhA6+78JfB+3DTPw==",
       "dev": true
     },
-    "@protobufjs/utf8": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/@protobufjs/utf8/-/utf8-1.1.0.tgz",
-      "integrity": "sha1-p3c2C1s5oaLlEG+OhY8v0tBgxXA="
-    },
     "@sindresorhus/is": {
       "version": "0.14.0",
       "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-0.14.0.tgz",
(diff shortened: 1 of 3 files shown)


dominguesgm:
Note: If the validation code runs calls to the IPFS API, they still need to be protected with a `pTimeout`.

Another approach could be to wrap the call to the validation function with a pTimeout to generically protect it from hanging. @terichadbourne any thoughts on this? It would mean having to come up with a generic error about not being able to find a specific `CID` during the validation.

terichadbourne:
I just tried using an infinite while loop and it froze my whole browser tab rather than returning this new error: 
image

I double-checked and I get the same response on the live version of the site, so it's not something introduced in this PR. The moving dots in the submit button freeze up almost immediately and my cursor has the disabled symbol on it when it happen.  Can we take a closer look at how the code is running and see if there's a different place we'd need to set a timeout on that? 

I do get this newly created error message if I invent my own CID to try to copy in this MFS lesson, but links aren't formatted properly in the error: 
image

I tried wrapping the `err.message` in [`marked`](https://marked.js.org/#/README.md#README.md) (the library that normally makes our `logDesc` markdown look nice but it got worse: 

```js
err.message = marked('Your code took too long to execute.... about [looping and iteration in JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Loops_and_iteration) ... challenge.')
```

image

Then I removed the mark and tried adding some non-link markdown instead to see if mark might be being used already but just not support links, but from the results it looks like it doesn't get used at all: 

```js
 err.message = 'Your _code_ **took** too long to execute.... challenge.'
```
image

I also edited a `logDesc` to include a markdown link and confirmed that it does appear as a link (although the contrast is poor): 
image


So next steps are: 
[ ] change the existing error message to be only about IPFS, not infinite loops
[ ] figure out where we can address infinite loops
[ ] figure out if there's a way we can make links work in error messages created here, rather than just those created in `logDesc` within a specific lesson's validation

dominguesgm:
You're right about the infinite loops, @terichadbourne , I guess I didn't think that part completely through. Thinking about it, I don't know if we have any way of protecting the user against infinite loops, unless they have some kind of asynchronous call in the middle of it. Since the browser is too busy executing the code inside the infinite loop, it will never finish and handle the timeout event.

To prevent this sort of stuff we would probably have to look into static loop termination checking, which is a whole can of worms we don't want to be opening right now.

In my latest commit I covered the other two points you mentioned: the error message and the markdown in the error messages.

terichadbourne:
@dominguesgm Sounds like you're feeling done with this one? Are you able to determine what's blocking the Travis CI build? Is it related to the `package-lock.json` conflict and is that one you could resolve? 

Do you want to merge this PR independently of Alex's, or have you discovered any ways in which they're interrelated so we should hold off?

dominguesgm:
I think it won't have any conflicts with Alex's PR and can be merged independently.

So far I've found the issue only happens when using `npm ci`and not with `npm install`. This is why it breaks on Travis and wasn't breaking on my machine. I'm trying to figure out exactly why this happens.

terichadbourne:
@dominguesgm I removed one more line from the error message that was also there because of the possibility of infinite loops and therefore isn't relevant. Other than that this looks good to me, so feel free to merge if you're good with the edit.

Issue #307, opened 2019-10-08
Epic 4B: Create or improve 3+ ProtoSchool tutorials based on IPFS Camp content


terichadbourne:
Please use this epic for relevant ProtoSchool-related tasks of the IPFS Docs WG. Due to filtering limitations, it lives in the ProtoSchool repo.

To add an issue from an the docs repo to this card, you'll need to remove any repo-specific filters so that this epic and your issue appear in ZenHub at the same time. 

You can see the [ProtoSchool Q4 OKRs](https://github.com/ProtoSchool/roadmap/blob/master/okrs/2019-q4.md) (which also include things on the community front unrelated to documentation/website) in our roadmap repo.

______

## 4B. Create or improve ProtoSchool tutorials using content adapted from IPFS Camp
<img src="https://img.shields.io/static/v1?label=Priority&message=P1&color=orange">

**_Find all issues associated with this epic [here](https://github.com/ipfs/docs/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen++%22epic+4b%22+)._**

### Deliverables
Tutorials finished and shipped

### Owners
- @terichadbourne & @dominguesgm 

### Context
Please see the Documentation & Developer UX Working Group's [Q4 2019 OKRs](https://docs.google.com/spreadsheets/d/1VeyiLvBdX_PrP394kU_lwkQZxfNwqMVX1f7K4ursSPM/edit#gid=1841105909) for information on how this epic fits into the team's larger goal structure for the quarter.

Issue #261, opened 2019-07-11
Create plan for use of IPFS Camp content in ProtoSchool


terichadbourne:
Review IPFS Camp content and determine what would fit well within ProtoSchool and whether it makes sense as new tutorials or additions to existing materials.

**Camp content to review** 
For some of these it will become quickly apparent that the content won't translate well to ProtoSchool, either due to the format or the conent (for example, vendor-specific content if there aren't extractable IPFS tidbits), but we'll at least note that quickly in this issue before moving on. Please check off below when the course has been reviewed and suggestions have been added to a new comment in this issue. 

- [x] Core A | Understanding how the InterPlanetary File System deals with Files | [Course Repo](https://github.com/ipfs/camp/tree/master/CORE_AND_ELECTIVE_COURSES/CORE_COURSE_A), Slides ([PDF](https://github.com/ipfs/camp/blob/master/CORE_AND_ELECTIVE_COURSES/CORE_COURSE_A/IPFS_Camp_Core_Course_A_Slides.pdf))| [Video](https://youtu.be/Z5zNPwMDYGg) |

- [x] Core B | Solving distributed networking problems with libp2p  | [Course Repo](https://github.com/ipfs/camp/tree/master/CORE_AND_ELECTIVE_COURSES/CORE_COURSE_B), Slides ([Google Slides](https://docs.google.com/presentation/d/1a_BjIM4ORQchnMNjOMO-wCDW2VS5nXeKXIbWSEqf9jY/edit?usp=sharing))| [Video](https://youtu.be/Q4IH5rWEO-E) |
- [x]  Core C | Developing Apps with IPFS API  |   [Course Repo](https://github.com/ipfs/camp/tree/master/CORE_AND_ELECTIVE_COURSES/CORE_COURSE_C), Slides ([PDF](https://github.com/ipfs/camp/files/3525750/IPFS.Camp.2019.-.Core.Course.C.-.Slide.Deck.pdf), [Google Slides](https://docs.google.com/presentation/d/1cbJD5j_jRpm3yJiE6hLIOzMn8iMhvzyci7Ut3NLjqxs/edit?usp=sharing))| [Video](https://youtu.be/6EHBOnFQJN4)|
[...]

PR #321, opened 2019-11-13
Add link to CID inspector when result is CID

### LINKED ISSUE #299: Feature: Link to CID inspector from lessons where result is a CID
We currently offer the option to view results in IPLD Explorer from many of our DAG lessons. 

image

Now that we have a [CID inspector](https://cid.ipfs.io), it would be cool to offer a similar option for lessons where the result is a CID. It looks like you can use the URL to load results for a specific ID, so for example, 
https://cid.ipfs.io/#QmYdGGwQXpaBGTVWXqMFVXUP2CZhtsV29jxPkRm54ArAdT pre-fills the search field and displays the results like so: 
image

We would presumably need a `resultIsCid` function to help us determine when we'd display this option. 

Obviously this tool will be most useful when we also take some of the content from @alanshaw @mikeal @Stebalien's IPFS Camp talk (Core Course A: Understanding How IPFS Deals with Files) and add it to ProtoSchool, so that we have content about the anatomy of a CID that can help people understand the tool, and we could then link to that tutorial from the tool itself.

terichadbourne:
Closes #299 

Partially addresses some of the suggestions for use of IPFS Camp content identified in #261, partially addressing epic #307.

When the result of a code challenge is a CID, a link to "View in CID Inspector" (which goes to a pre-filled URL in https://cid.ipfs.io/) now appears alongside the link to "View in IPLD Explorer."

I'd love feedback on the ideal layout since both options would always be present at the same time. Current state of wrapping shown below, although few people will be attempting to code on narrower screens.
image
image
image

### DIFF
--- a/src/components/Output.vue
+++ b/src/components/Output.vue
@@ -46,7 +46,7 @@ export default {
     inspectCidUrl: function () {
       let cid = this.output.test && this.output.test.cid && this.output.test.cid.toBaseEncodedString()
       cid = cid || ''
-      return `https://cid.ipfs.io/#/${cid}`
+      return `https://cid.ipfs.io/#${cid}`
     }
   }
 }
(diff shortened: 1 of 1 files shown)


terichadbourne:
Nice catch. Thanks!

PR #326, opened 2019-12-06
Add rendering tests

### LINKED ISSUE #323: Create Cypress tests for file upload
As it currently stands, we use cypress as our testing framework, for which we have tests for every lesson in tutorial 2 and 3.

There is an issue with building tests for the other tutorials, which make use of file upload, which from what I know, Cypress is not able to mock. *However*, with the way we store uploaded files in an array in the `window` object, I believe we can dynamically create a file for testing purposes. See [this link](https://stackoverflow.com/questions/27159179/how-to-convert-blob-to-file-in-javascript), which may be useful.

The biggest driver from this issue, in my opinion, comes from the fact that every time some core functionality is changed - for example how we're changing `Lesson.vue` in #301 - we have to manually run error cases to see if the user code output and validation output are being handled correctly in all cases, which is time consuming and should be automated.

terichadbourne:
A new Cypress test loops through all tutorials in `tutorials.json` and ensures that each individual lesson and resources page is listed on the tutorial page and is able to render. This will prevent further errors with routes as recently discovered and fixed in #325 and noted in [this comment](https://github.com/ProtoSchool/protoschool.github.io/issues/323#issuecomment-562629335) on issue #323 (raised by @dominguesgm). This new test is run on every tutorial and lesson (including resources pages) regardless of lesson type.

The existing test built by @olizilla to check that all solution code actually passes validation has been updated slightly to: 
- reflect our new mandatory status for resources pages
- allow simpler updates for which tutorials should have this test run (just add their IDs to the relevant array, which will be looped through)

Note that the current solution for that requires that every single lesson (except the resources page) in a tutorial be a standard coding challenge. 

This PR does NOT close #323 because it doesn't address the issue of testing multiple choice or file upload lessons or ensuring progression works through text only lessons. Once we make those updates we'll likely be able to do more nesting of tests, as in "go to next lesson, if it's X type of lesson, test it like this, repeat." However, I think it stands on its own and can be merged once reviewed, with that issue left open for future improvements.

**A sampling of successfully caught author errors:** 

- If an author screws up the routing in `main.js` as was done previously, the rendering test will fail because the lesson will be listed from the tutorial landing page but clicking it will lead to a Page Not Found: 
```js
  { path: '/mutable-file-system/09', component: T0004L09 },
  { path: '/mutable-file-system/10', component: T0004L10 },
  { path: '/mutable-file-system/01', component: T0004L11 },
```
image
image

- If an author forgets to include a resources list in `tutorials.json`, Cypress tests will now fail in two places: 

(a) The rendering test will fail because there will be no line item for a resources page in the tutorial landing page:

image
image

(b) The lesson-passing test will fail because the last normal lesson in the tutorial will generate a next button to a /resources URL. At the moment it will actually render a page but with no useful content. 

image
image

We could call it a bug that we're generating content-less resources pages in the case of this error, but the test will prevent us from committing the error, so it doesn't necessarily need to be fixed. (I caused this error intentionally for testing; all tutorials DO have resources content.)

### DIFF
--- a/cypress/integration/tutorials.spec.js
+++ b/cypress/integration/tutorials.spec.js
@@ -2,12 +2,26 @@
 
 import tutorials from '../../src/static/tutorials.json'
 
+describe('show 0001', function () {
+  renderAllLessonsInTutorial('0001')
+})
+
 describe('📝 0002', function () {
+  renderAllLessonsInTutorial('0002')
   viewSolutionsAndSubmitAll({ tutorialId: '0002', lessonCount: 3 })
 })
 
 describe('📝 0003', function () {
+  renderAllLessonsInTutorial('0003')
   viewSolutionsAndSubmitAll({ tutorialId: '0003', lessonCount: 7 })
+})
+
+describe('show 0004', function () {
+  renderAllLessonsInTutorial('0004')
+})
+
+describe('show 0005', function () {
+  renderAllLessonsInTutorial('0005')
 })
 
 function viewSolutionsAndSubmitAll ({ tutorialId, lessonCount, hasResources = true }) {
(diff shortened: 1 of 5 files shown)


dominguesgm:
Found a problem which was breaking the `should find resources and navigate to tutorials` test and commited the changes to fix it.

Other than that, it looks good to me 👍

terichadbourne:
@dominguesgm Thanks for testing! Could you tell me more about the circumstances that caused the problem with `contains` but not with `get`? The tests were passing and failing under the right circumstances on my machine, and my impression is that `contains` returns the first matching DOM element whereas `get` returns them all. In the spot that you edited it seems like we only need to know that at least one element with the given data attribute appears on the page, so it seems like either should do the trick.  I'd love to hear more about what wasn't working for you.

dominguesgm:
That's odd because those two new tests were always failing on my machine. From my limited cypress experience (which means I might be incorrect 😅) :

* *get* returns an html element given a selector,
* *contains* returns an html element given [a selector and] textual content (the selector is optional)

[Relevant documentation.](https://docs.cypress.io/api/commands/contains.html#Syntax).

What we had before was a `contains` call with a selector and no text content, which leads to an error. I'm honestly a bit unsure about how it was working before.

Issue #120, opened 2019-01-08
Provide better error message when wrong CID returned (Basics lesson 2)


lanzafame:
Have a question or suggestion regarding a specific ProtoSchool lesson? Please use this template to share it!

**URL of the lesson that's confusing:**
https://proto.school/#/basics/02

**What's confusing about this lesson?**
The error message returned when the wrong cid is returned. It isn't readable (data is return as bytes), and the message also get truncated.
See: 
![1546913603](https://user-images.githubusercontent.com/5924712/50807040-d3956d00-1344-11e9-9f8e-d7ba22e3ac70.png)

mikeal:
@terichadbourne should we create a new issue detailing the *fix* for this problem, or should we rename the issue here and start detailing the fix in-line?

mikeal:
In *most* of the tutorials we have code that checks if there is a CID and if it is incorrect it produces a different error message that just not returning a CID at all, or returning some other incorrect value. The problem is, we didn't update this after the CID API changes, it used to show the base encoded string for the CID but now it just tries to print the CID in several places. We just need to go in and clean up those error messages to print the base encoded CID.

terichadbourne:
I just tested this output again and it's no longer truncated on the live site but does still return data in the format shown above (objects, not strings). 

Just submitted a potential fix in #328 but would love feedback as to whether it's a reasonable approach before proceeding.

terichadbourne:
previously closed via #328

PR #328, opened 2019-12-10
Display CID strings, not objects

terichadbourne:
@mikeal this is a first stab at a fix to the issue @lanzafame raised in #120 where we were printing CID objects instead of just CID strings. 

This works and displays both CIDs as strings in the output when they don't match, but in order to make it recognize a correct answer in this approach I had to switch to loose equality to make the result (a CID which comes as an object) match the correct CID (expressed as a string). Is it safe to do it this way or would you recommend some kind of actual conversion? 

I'll circle around and look for other cases of this problem to fix after confirming whether this approach feels reasonable.

### DIFF
--- a/src/components/Explorer.vue
+++ b/src/components/Explorer.vue
@@ -8,8 +8,7 @@
 <script>
 export default {
   data: self => {
-    let cid = self.$attrs.cid
-    cid = cid.toBaseEncodedString ? cid.toBaseEncodedString() : cid
+    let cid = self.$attrs.cid.toString()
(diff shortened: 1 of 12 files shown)


terichadbourne:
Rather than doing `result == correctHash` (knowing `result` is an object that can be coerced to a string and `correctHash` is already a string, @mikeal proposed doing `result.toString() === correctHash`. `toBaseEncodedString` is the old, but still there, API, `.toString()` is nicer and newer ;)

I've updated `toBaseEncodedString()` to `toString()` throughout the codebase and removed an old reference where we send people to learn about the `toBaseEncodedString` method. Changes to specific lessons are all in the Basics and Blog tutorials, but Explorer.vue and Output.vue also include updates.

PR #334, opened 2019-12-20
feat: include code in issue

### LINKED ISSUE #315: Feature: Submit code snapshot after unexpected error while validating
When a user submits code in a Coding Lesson, and our validation returns a generic error message, the user gets no additional feedback about their answer other than "it's incorrect", without any hint of what they should try to do differently.

A way to make use of this scenario to improve the lessons and their validation would be a link/submission form which would create a Github issue, with a predefined template for this very purpose, containing a copy of the code of the user which led to the generic error feedback.

This link/form would only appear after a solution submission which resulted in a generic error feedback.

dominguesgm:
Link for validation failure issue now contains code automatically.

This was done for the two possible outputs that may results from issues with our validation.

This closes #315.


src/components/Lesson.vue
```diff
@@ -292,6 +290,21 @@ export default {
...
+      return newGithubIssueUrl({
+        user: 'ProtoSchool',
+        repo: 'protoschool.github.io',
+        title: `Validation Error: ${this.tutorialShortname} - Lesson ${this.lessonNumber} (${this.lessonTitle})`,
+        labels: ['lesson-feedback', 'validation-error'],
+        body: `If you submitted code for a lesson and received feedback indicating a "validation error", you may have uncovered a bug in our lesson validation code. ${
+          validationTimeout
+            ? 'In this case, the error experienced was caused by an issue that caused the validation function to hang.'
+            : 'This may be a result of a validation case we forgot to cover.'
+        } Please use this template to help us diagnose the problem.
```
terichadbourne:
```suggestion
        } Thank you for submitting your feedback to help us diagnose the problem.
```

terichadbourne:
I got this error by changing a few characters in the correct hash, trying to recreate a condition in which the code would time out: 

image

But on closer review it looks like this is a totally separate error, which doesn't currently load with an option for reporting it. Is that correct?

The text changes I suggested above were created without being able to preview the actual issue creation during testing.

Could you please suggest specific ways to reproduce the two errors you're addressing in this PR to make it easier for folks to test that the correct error messages are appearing and the issue template is populating correctly? (Both what code to submit as a fake user and what changes to make under the head during testing, such as removing some necessary validation code.) 

Just confirming you're now creating an issue from scratch rather than using the template created in #324 (which only exists in `code` and not in the branch you're merging into)?

dominguesgm:
The error you came across is correct as it is a user code error and not a validation error. In order to replicate the errors that would lead to the issue creation link, you have to modify the validation code of a lesson.

For example, if you want to replicate the error for when a validation fails to find the matching scenario, you can simply replace the whole body of a lesson's validation method with a `return null`.

Replicating the other error message (which occurs when the validation throws an error), follows a similar approach, but instead of `return null` you need to insert a `throw new Error()` line.

terichadbourne:
Thanks for the explanation @dominguesgm! I was able to replicate the behavior using those clues and it's looking great.  

I made a few edits to the body of the error messages and the pre-populated issues. The changes to the issues themselves are mostly to limit the chance that the user will delete any information we actually need. 

If you don't have any concerns with my changes, please go ahead and merge this into the infamous Alex branch and then we'll ask again for a review on that one. Thanks!

dominguesgm:
Looks good, I can't approve my own but I'll merge it 😄

PR #336, opened 2020-01-06
Add video from IPFS Camp Core Course D to resources pages

### LINKED ISSUE #333: Add Core Course D to resources pages
_This issue is a result of evaluation of IPFS Camp content for inclusion in ProtoSchool. See #261._

Link to the [video](https://youtu.be/fLUq0RkiTBA) of IPFS Camp 2019 **Core Course D: The lifecycle of data in the DWeb** from resources pages  throughout the site. Add to: 

- [x]       Decentralized Data Structures tutorials (great more advanced follow-up on the concept of sharing data without making learners code)
- [x]       All of the coding challenges as a "you've added a file to IPFS, now what?"

terichadbourne:
Closes #333 by adding a link to the video from IPFS Camp Core Course D (Lifecycle of Data in the DWeb) to all existing resources pages, with slightly different descriptions depending what the user has already learned. 

Also adds markdown functionality for resource descriptions by importing `marked`. Because the text now pulls in as inner HTML, as a child of the div we created directly, I found I need to create a new custom class to adjust the spacing on the `p` element that pulls in (to keep the description directly under the resource title). As far as I know I can't use Tachyons directly to affect the child of the element I have access to.

### DIFF
--- a/src/components/Resources.vue
+++ b/src/components/Resources.vue
@@ -10,7 +10,7 @@
         <a class='b blue link' :href='item.link' target='_blank'>{{item.title}}</a>
         <span class='ml2 ph2 bg-navy br-pill white f7'>{{item.type}}</span>
         <span
-          v-if='item.link.includes('/proto.school/')'
+          v-if='item.link.includes("/proto.school/")'
           class='ml2 ph2 bg-aqua br-pill white f7'
         >ProtoSchool</span>
       </p>
@@ -35,7 +35,7 @@ export default {
 </script>
 
 <style>
-.resource-desc > p {
+.resource-desc p {
   margin-top: 0;
 }
 </style>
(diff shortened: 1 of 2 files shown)


src/components/Resources.vue
```diff
@@ -7,15 +7,28 @@
...
 
 <script>
+import marked from 'marked'
+
 export default {
   props: {
     data: Array
+  },
+  methods: {
+    parse(description) {
```
zebateira: linter issue: `error: Missing space before function parentheses`

src/components/Resources.vue
```diff
@@ -7,15 +7,28 @@
...
+      return marked(description || '')
+    }
   }
 }
 </script>
+
+<style>
+.resource-desc p {
```
zebateira: should it be more specific? Something like:`.resource-desc > p`.

Issue #317, opened 2019-10-31
Improve button contrast


lidel:
> @ericronne  mentioned this problem in WebUI and now I cannot unsee it everywhere :eyes: :sweat_drops: 

### Problem

Buttons have low contrast and look "disabled" and literally look the same when "disabled" and "enabled":

- Disabled
  - ![off-2019-10-31--15-23-14](https://user-images.githubusercontent.com/157609/67955565-12a5ef00-fbf3-11e9-86d3-2fe7d9a627e6.png)

- Enabled + hover
  - ![on-2019-10-31--15-23-33](https://user-images.githubusercontent.com/157609/67955564-120d5880-fbf3-11e9-9c9a-44c1e0324fb4.png)
  - ![2019-10-31--15-47-03](https://user-images.githubusercontent.com/157609/67957540-11c28c80-fbf6-11e9-9fd1-d2702b792de4.png)

### Proposed change

It would be useful to change background of "enabled" buttons to something like `bg-navy` from [ipfs-css](https://github.com/ipfs-shipyard/ipfs-css):

![Screenshot_2019-10-31 ProtoSchool](https://user-images.githubusercontent.com/157609/67957846-84cc0300-fbf6-11e9-8994-2f716b35ce0e.png)

terichadbourne:
@lidel Do you remember where you were / what you were doing when you got a disabled button that looks like this? We do have at least one case where we disable it (when you haven't yet uploaded files, but that case contains a hover state with a message that makes it abundantly clear why you can't click. 

(I'm open to color changes also, just want to make sure we also know when/why it's happening in case there's other stuff to address.

ericronne:
Fyi disabled buttons don't have to meet the 4.5:1 guideline mark.
https://www.w3.org/TR/WCAG21/#contrast-minimum

lidel:
@terichadbourne it was `/#/regular-files-api/05` (i cropped the screenshot, below is full version of disabled button): 

![2019-11-01--13-34-45](https://user-images.githubusercontent.com/157609/68025174-77715000-fcac-11e9-98eb-47c02ace8cee.png)


What I meant that disabled button looks ok, the problem is contrast of active ones :)

ericronne:
I'd recommend that the inactive button color be a tint of the active button, rather than a different color altogether. So if the submit button will be blue when active, it could look something like this when inactive (alphaed back blue for the fill, alphaed back white text):
[...]

PR #342, opened 2020-01-09
chore: improved button disabled status

dominguesgm:
Updated disabled button styles to match @ericronne's recommendation in #317.

Screenshot 2020-01-09 at 15 26 48

One thing to note is that as the button is disabled in the loading state, it will show up with the new disabled style. We need to discuss whether this is appropriate or not @terichadbourne @zebateira.

Screenshot 2020-01-09 at 16 33 26

### DIFF
--- a/src/components/Button.vue
+++ b/src/components/Button.vue
@@ -33,8 +33,10 @@ button:active {
 }
 
 button:disabled {
-  opacity: 0.8;
+  opacity: 0.95;
   box-shadow: initial;
   cursor: default;
+  color: #b9c4cd;
+  background-color: #8498a6;
 }
 </style>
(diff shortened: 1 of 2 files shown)


terichadbourne:
Thanks @dominguesgm! I believe @ericronne's suggestion for the disabled color scheme was meant to coordinate with @lidel's suggestion of using  `bg-navy` from `ipfs-css` for the active versions instead of the aqua we have now (for better color contrast). Mind updating that as well?

dominguesgm:
Sure, just tried it. Let me know if you prefer this look.

Screenshot 2020-01-10 at 10 14 25

zebateira:
@dominguesgm good to go. just commit the changes and we can merge 🎉

PR #346, opened 2020-01-16
Feat/add tutorial types

### LINKED ISSUE #310: Feature: Label for coding v non-coding tutorials
It would be great to have a visual indicator that quickly identifies whether the tutorial requires coding skills from three spots: 
- the tutorial card in the featured tutorials section of the [homepage](https://proto.school/#/) 
- the tutorial's listing within the [tutorials page](https://proto.school/#/tutorials) 
- the tutorial's own landing page (eg https://proto.school/#/data-structures) 

Right now we have mostly tutorials where a lack of JavaScript skills will be a blocker, plus one that's reading-only, and the completion rates suggest that there's a decent drop-off once people realize coding is required. Let's give people clues sooner! 

Maybe these would be an icon for a book (or something else to indicate reading) and an icon for code or JavaScript? (At the moment and for the foreseeable future there will be no languages other than JS used, but it is important for coders of all sorts to know that JavaScript is required.) 

There will be multiple-choice lessons mixed into some of the tutorials moving forward, but since these could be found in both otherwise-reading-based and otherwise-coding-based tutorials, I think the differentiator we need for a whole tutorial is just whether or not it requires coding.

terichadbourne:
Closes #310


terichadbourne:
I just opened an issue in `design-shop` to request icons we can sub in here ( https://github.com/protocol/design-shop/issues/256), and we won't be able to merge until we have those. 

I'm still working on: 
- creating a single component to generate these logos in the various places they're displayed
- fixing an apparent logic issue: the tutorial type is calculating to a different answer on the tutorials page than it does on an individual tutorial's landing page, despite those being rendered by the same component

DEVELOPING_TUTORIALS.md
```diff
@@ -485,16 +485,25 @@ In `static/tutorials.json`, add a new key for your tutorial and fill in the appr
   "title": "Your short tutorial title",
   "description": "Your tutorial description",
   "lessons": [
-    "Title of 1st lesson",
-    "Title of 2nd lesson",
-    "Title of 3rd lesson"
+    {
+      "title": "Title of a coding lesson (either standard or requiring file upload)",
+      "types": ["code"]
```
zebateira: thinking through this, is it necessary to be an array since we are only showing one icon?

src/pages/Home.vue
```diff
@@ -17,7 +17,13 @@
         <template v-for="(tutorial, index) in featuredTutorials">
           <div class="bg-aqua br4 pa3 mb3 tutorial-tile" :key="index">
             <router-link :to="tutorial.url">
+              <div class="flex flex-row justify-between ">
               <h3 class="ma0 f3 fw7 navy">{{tutorial.title}}</h3>
+              <img
```
zebateira: still need to use the component here

DEVELOPING_TUTORIALS.md
terichadbourne: we had talked about adding `file-upload` here in the future, which wouldn't affect the icon but would help Cypress figure out what test to run

zebateira: Ah yes.
Looking back at it seems to be a bit confusing though.
I think that we can assume `file-upload` lessons also include code, meaning we can simply either `code` or `file-upload` type on a lesson, which allows to use it as a visual hint in the UI and also cypress will be able to do the proper testing.

So effectively, we just need the property `type` and its values can be `text`, `multiple-choice`, `code` or `file-upload`.

Sounds good?

terichadbourne: yep - i can refactor to make this work

terichadbourne: done in https://github.com/ProtoSchool/protoschool.github.io/pull/346/commits/852f2c7fa586b3815d995c9c386c4eeef0e91fe6 - note that this will also need to be fixed in your new multiple-choice tutorial

zebateira:
# Summary of the PR

## Code changes

- Added new `ToggleButton` adapted from [here](https://github.com/euvl/vue-js-toggle-button)
- Added new `TypeIcon` component to render an icon that represents the type of a lesson or tutorial
  - To support this, we made the lesson string in the `tutorials.json` file an object with `{ title, type }` properties
  - The tutorials type is automatically computed from the types of the lessons
  - Added new svg loader `vue-cli-plugin-svg` to support rendering icons inline to be able to change their color
- Refactored tutorials grid in the homepage into a new component `TutorialsGrid`
- Added [`v-tooltip`](https://github.com/Akryum/v-tooltip)


## UI Changes

### Homepage: tutorial card UI tweak

image

Each tutorial now includes the type with an icon and a tooltip.:

image

### Tutorials page: added type icons and coding tutorials toggle

image

@terichadbourne I made the icons in the lesson link green, let me know if you prefer the lesson title color. My reasoning was that we want to call attention to the lesson title, not the elements in green (lesson x, icon, etc)

src/components/TypeIcon.vue
```diff
@@ -0,0 +1,54 @@
...
+      const isLesson = !!this.lessonId
+      const type = this.lessonId
+        ? getLessonType(this.tutorialId, this.lessonId)
+        : getTutorialType(this.tutorialId)
+
+      switch (type) {
+        case 'code':
+          return { icon: 'code', alt: 'JavaScript code challenge' + (!isLesson ? 's' : '') }
+        case 'multiple-choice':
+          return { icon: 'multipe-choice', alt: 'Multiple choice quiz' + (!isLesson ? 'es' : '') }
```
terichadbourne:
```suggestion
          return { icon: 'multipe-choice', alt: 'Multiple-choice quiz' + (!isLesson ? 'zes' : '') }
```

terichadbourne:
@zebateira Thanks for the summary and the latest updates. Just a couple of styling suggestions:

I'd prefer for the tutorial description here to always start the same distance from the header, as opposed to leaving various-sized gaps. (I'm fine with extra blank space at the bottom to make the cards in each row be the same height as each other.)

image

I'm okay with using the green icon here for the reason you mention:
image

... but I feel like under the same theory we should probably then make it green here to match:
image

Thoughts?

zebateira:
@terichadbourne 

> ... but I feel like under the same theory we should probably then make it green here to match:

yes, makes sense. will update it

zebateira:
@terichadbourne 

> I'd prefer for the tutorial description here to always start the same distance from the header, as opposed to leaving various-sized gaps. (I'm fine with extra blank space at the bottom to make the cards in each row be the same height as each other.)

Agreed 👍

terichadbourne:
@zebateira These changes look great, thanks! I don't see any merge conflicts and am comfortable with merging if you are.

PR #353, opened 2020-01-26
Make "Decentralized Data Structures" first tutorial

carlbordum:
Hello, I just started using https://proto.school/ and I think it is awesome! I will make PRs for the few things that are different than I expected. Repeat of commit message:
```
This patch changes the order of the tutorials as presented
https://proto.school/#/tutorials. The reason for this is that the MFS
and Files API tutorials reference it. Here are the pages referencing the
data structure tutorial:
https://proto.school/#/mutable-file-system/01
https://proto.school/#/mutable-file-system/06
https://proto.school/#/regular-files-api/01
```


terichadbourne:
Totally agree. Thanks @carlbordum!

PR #355, opened 2020-01-26
Fix markdown codeblock rendering issue

carlbordum:
This seemingly innocent newline makes sure the code block renders correctly.
The file is rendered here: https://proto.school/#/regular-files-api/07


zebateira:
again, great catch @carlbordum 🙏

PR #356, opened 2020-01-26
Tutorial 5, Lesson 5 exercise validation fix

carlbordum:
Before this patch, the exercise would fail - even with the solution code - when only one file was uploaded, as the `resultingFiles.length` wasn't guarded by `resultingFiles.length !== uploadedFiles.length`.

To reproduce:
  (1) upload one file
  (2) copy solution code
  (3) click "Submit"


zebateira:
thank you @carlbordum 🙏 
Good catch.

PR #359, opened 2020-01-27
Add new tutorial - "Anatomy of a CID"

### LINKED ISSUE #329: New Tutorial: Anatomy of a CID
_This issue is a result of evaluation of IPFS Camp content for inclusion in ProtoSchool. See #261._

We'd like to build a new course on the anatomy of a CID - including details on the intricacies of hashing - based on @alanshaw's Core Course A: Understanding How IPFS Deals with Files from IPFS Camp 2019 ( [video](https://youtu.be/Z5zNPwMDYGg) | [slides](https://github.com/ipfs/camp/blob/master/CORE_AND_ELECTIVE_COURSES/CORE_COURSE_A/IPFS_Camp_Core_Course_A_Slides.pdf) | [course repo](https://github.com/ipfs/camp/tree/master/CORE_AND_ELECTIVE_COURSES/CORE_COURSE_A)).

This course would most likely be a fit for a multiple-choice lesson format and would incorporate the great visuals included in the original presentation. It would also highlight the CID Inspector tool built by @olizilla and already incorporated throughout our lessons (#299). 

Further detail and outline to follow.

### LINKED ISSUE #364: Feature: Enable markdown on multiple-choice answer text
In order have more pleasant experience on multiple-choice lessons, we should enable markdown to the answers text. This would allow to highlight some text, with bold italic, etc.

Ref: https://github.com/ProtoSchool/protoschool.github.io/pull/359#pullrequestreview-353151422

image

zebateira:
Closes #329
Closes #364


carlbordum:
Hi @zebateira!

I just worked through this tutorial and it is great! I encountered no issues. I have submitted #361 that changes stuff a bit when I had to read it twice or noticed an inconsistency.

On lesson 03, IPLD is mentioned with no further explanation or references. Is this good?

The infographics are very nice, but I think they may have accessibility issues for people who suffer from reduced vision, color blindness or blindness. This is probably out of scope for this pull request, but I thought I would mention this.

Have a nice day and thanks for this tutorial :- )

zebateira:
> I just worked through this tutorial and it is great! I encountered no issues. I have submitted #361 that changes stuff a bit when I had to read it twice or noticed an inconsistency.

Merged 👌 

> On lesson 03, IPLD is mentioned with no further explanation or references. Is this good?

Hm, indeed, maybe it would be nice to have a short sentence and reference link to contextualize it better.

> The infographics are very nice, but I think they may have accessibility issues for people who suffer from reduced vision, color blindness or blindness. This is probably out of scope for this pull request, but I thought I would mention this.

The images should have an `alt` attribute. Does it need something more?

Thank you again for your contributions 🙏

carlbordum:
Honestly, I don't really know much about accessibility (will read up, it's important!). When I did my review, at first I didn't rebuild, so the pictures didn't load which made me realise the difference. So that's why I mentioned it

zebateira:
> Honestly, I don't really know much about accessibility (will read up, it's important!). When I did my review, at first I didn't rebuild, so the pictures didn't load which made me realise the difference. So that's why I mentioned it

I see, but in that case you should've seen the alt text displayed instead, no?

carlbordum:
I see it after rebuilding :-)

zebateira:
> I see it after rebuilding :-)

Ah ok.
@terichadbourne let me know what you think, but I think the `alt` text should be enough.

src/tutorials/0006/03.vue
```diff
@@ -0,0 +1,42 @@
...
+const question = 'How can we identify the encoding used on specific data when using CIDs?'
+
+const choices = [
+  {
+    answer: 'CIDs have a multicodec prefix that specifies in which encoding is the data represented',
+    correct: true,
+    feedback: 'That\'s right!'
+  },
+  {
+    answer: 'All encodings make it safe to try and check the first bytes of the data to figure out which encoding was used',
```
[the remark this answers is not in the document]
zebateira: This is the problem we are trying to solve: we can stop making assumptions (incorrect ones, unsafe ones) about the encoding used, and actually just read the multicodec identifier from the CID to understand what was the encoding used.

src/tutorials/0006/05.md
```diff
@@ -0,0 +1,25 @@
...
+
+A detail about this prefix, is that it is **only used in the string form of the CID**.
+
+So let's see two examples of CIDs in the string form:
+
+![String form examples](tutorial-assets/T0006L05-string-form.jpg)
+
+The first one is a `CIDv0` because it starts with `Qm...`. All Hashes that start with `Qm` can safely be interpreted as `base58btc` as a CID of version 0.
+
+The second one starts with `bafy...` and the first character `b` is the base encoding prefix indentifier because, usually, most implementations of IPFS by default encode the CIDs using `base32`, which its identifier character is `b`. For the complete list of identifiers of `multibase` see [this table](https://github.com/multiformats/multibase/blob/master/multibase.csv).
```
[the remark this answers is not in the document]
zebateira: There are some variations, but `bafy...` CIDs are identifiable as CIDv1.
CIDv2 would not start with `bafy...` because of the version prefix if I'm not mistaken.

zebateira:
@terichadbourne 

I pushed some edits myself after reviewing your review (accidentally stumbled upon several repeated 'the' `the the` so that's included as well). Let me know if it makes sense.

> Throughout the tutorial, it feels odd to me when there are references to "which encoding was used" instead of "which encoding method" or "which type of encoding." I don't know if that's a genuine concern or is a matter of my technical vocabulary not being up to snuff. I haven't edited these spots because I'm not confident enough to do it, but maybe @alanshaw can weigh in when he reviews this.

Good question. Encoding refers to the representation of the data, that's why I didn't use "method" or "type, but maybe "type" could be a valid one. @alanshaw ?

> It also feels really awkward that we need to keep referring to CIDs of certain versions - essentially "a Version 0 CID" would be "a CIDv0 CID," which is repetitive. Maybe there are spots where we should replace CID with hash to make it smoother. For now I've just replaced "version zero" with "Version 0," etc.

Yes, this is much better 👌 

> The resources page for the tutorial isn't being checked off as complete in the table of contents after I visit it, and I have the same issue with some but not all of the other tutorials on this branch. May be a pre-existing problem based on lesson type?

Will take a look at this 👍 

> I don't recall if there was a reason we couldn't use marked in the multiple choice answers. It would be ideal to allow the use of code blocks with `` here but I quickly tried it and it interpreted them literally as text.

Yeah, we should enable this if we can. Issue created here: https://github.com/ProtoSchool/protoschool.github.io/issues/364

src/tutorials/0006/02.md
```diff
@@ -0,0 +1,35 @@
+By using a cryptographic algorithm to generate the hash, we guarantee that we can identify a file with a single hash.
+
+Occassionally, a hashing algorithm may stop being compliant with the previously stated rules, as has already happened with `sha1`. With time, other algorithms may prove to be insufficient for content addressing in IPFS. For this reason, and in order to support multiple cryptographic algorithms, **we need to be able to know which algorithm was used to generate the hash** of specific content.
```
alanshaw: There were called characteristics earlier, so using the same word here for consistency. Also "stop being compliant" suggests it has a choice in the matter 😂.

```suggestion
Occassionally, a hashing algorithm may be proven to be insecure, meaning it no longer complies with the characteristics that we defined earlier. It has already happened with `sha1`. With time, other algorithms may prove to be insufficient for content addressing in IPFS. For this reason, and in order to support multiple cryptographic algorithms, **we need to be able to know which algorithm was used to generate the hash** of specific content.
```

src/tutorials/0006/02.md
```diff
@@ -0,0 +1,35 @@
...
+
+To allow representation of a CID as a compact string (instead of plain binary), IPFS uses `base58btc` encoding to get the following:
+
+`QmY7Yh4UquoXHLPFo2XbhXkhBvFoPwmQUSa92pxnxjQuPU`
+
+So with this full specification, the first version of IPFS CIDs was born! This first version is called Version 0 (`CIDv0`), and the initial `Qm...` characters are easily spotted on its CIDs.
+
+However, with time, doubts about whether this format would be sufficient started to arise:
+
+ - How do we know the encoding of the data?
```
alanshaw: I think this could do with a little more explanation. I see it is explained further below but there's no mention of encoded data prior to this sentence.

src/tutorials/0006/02.md
```diff
@@ -0,0 +1,35 @@
...
+ - How do we know the encoding of the string representation of the hash? Will we always be using `base58btc`?
```
alanshaw:
```suggestion
 - How do we know the encoding of the string representation of the CID? Will we always be using `base58btc`?
```

src/tutorials/0006/02.vue
```diff
@@ -0,0 +1,42 @@
...
+import MultipleChoiceLesson from '../../components/MultipleChoiceLesson'
+import text from './02.md'
+
+const question = 'How do CIDs support multiple cryptographic algorithms?'
+
+const choices = [
+  {
+    answer: 'The length of the hash tells us which cryptographic algorithm was used. If the length is "256" then the algorithm used was "sha2-256". If the length is "512" then the algorithm used was "sha2-512."',
+    correct: false,
+    feedback: "It's true that the length of a `sha-256` hash is 256, but this isn't enough information to tell us what hashing algorithm was used. Check the information above and try again."
```
alanshaw:
```suggestion
    feedback: "It's true that the length of a `sha2-256` hash is 256, but this isn't enough information to tell us what hashing algorithm was used. Check the information above and try again."
```

src/tutorials/0006/03.md
```diff
@@ -0,0 +1,15 @@
...
+
+To solve this issue, `CIDv1` introduces another prefix that uniquely identifies the encoding used.
+
+## Multicodec prefix
+
+The **multicodec prefix** indicates which encoding was used on the data.
+
+![Multicodec Prefix](tutorial-assets/T0006L03-multicodec.jpg)
+
+In the example above we see how data encoded with the codec `dag-pb` would be represented in our CID.
```
alanshaw: Ok so, multicodec is just a numeric identifier. They are used for many different purposes. In multihash, the "type" is a multicodec. The multicodec we're talking about here is an "IPLD format". So it's not really correct to title this "Multicodec prefix", although that's techinically what it is, it isn't specific enough. It's like going to see Muse at a gig and the banner outside the venue says "Live tonight: Rock Band" not "Live tonight: Muse".

src/tutorials/0006/03.vue
```diff
@@ -0,0 +1,42 @@
...
+    correct: false,
+    feedback: 'Hmm, not always. Please check the information above'
+  },
+  {
+    answer: 'We can\'t know for sure, but the multicodec prefix included in the CID is a hint that is not always guaranteed to be right.',
```
alanshaw:
```suggestion
    answer: 'We can\'t know for sure, but the IPLD format multicodec included in the CID is a hint that is not always guaranteed to be right.',
```

src/tutorials/0006/03.vue
```diff
@@ -0,0 +1,42 @@
...
+    correct: false,
```
alanshaw: This is actually true, there's no guarantee when you get the data that it was encoded using the IPLD format in the CID.

src/tutorials/0006/03.vue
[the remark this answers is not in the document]
alanshaw: I'm not sure I understand this either.

src/tutorials/0006/04.md
```diff
@@ -0,0 +1,15 @@
+So now our Version 1 CID contains the following fields:
+
+`<multicodec><multihash>`
```
alanshaw: You might want to start with multihash expanded out as `<mh-algo><mh-len><mh-hash>` before shortening to `<multihash>` as a reminder.

alanshaw:
```suggestion
`<ipld-format><multihash>`
```

src/tutorials/0006/04.md
```diff
@@ -0,0 +1,15 @@
...
+
+But, if you remember from the previous lessons, Version 0 CIDs only contain the `<multihash>` part, so how can we distinguish between different versions of CIDs? You guessed it, more prefixes!
+
+![Version Prefix](tutorial-assets/T0006L04-version-prefix.jpg)
+
+So now our CID looks like the following:
+
+`<cid-version><multicodec><multihash>`
+
+The `<cid-version>` represents the version of the CID. In the future, it will increment to 2, 3, 4, etc.
```
alanshaw: Please no!

```suggestion
The `<cid-version>` represents the version of the CID (0 or 1).
```

src/tutorials/0006/04.vue
```diff
@@ -0,0 +1,42 @@
...
+
+<script>
+import MultipleChoiceLesson from '../../components/MultipleChoiceLesson'
+import text from './04.md'
+
+const question = 'Which version of CIDs include the actual version number of the CID?'
+
+const choices = [
+  {
+    answer: 'None. The version number is inferred by detecting whether the multicodec prefix is present.',
```
alanshaw:
```suggestion
    answer: 'None. The version number is inferred by detecting whether the IPLD format multicodec is present.',
```

src/tutorials/0006/05.vue
```diff
@@ -0,0 +1,42 @@
...
+        :question="question"
+        :choices="choices" />
+    </div>
+</template>
+
+<script>
+import MultipleChoiceLesson from '../../components/MultipleChoiceLesson'
+import text from './05.md'
+
+const question = 'How can we know the base encoding of the hash?'
```
alanshaw:
```suggestion
const question = 'How can we know the base encoding of a CID?'
```

src/tutorials/0006/05.vue
```diff
@@ -0,0 +1,42 @@
...
+
+const choices = [
+  {
+    answer: 'We have to guess it, but it\'s easy because most implementations of IPFS encode CIDs with "base32."',
```
alanshaw:
```suggestion
    answer: 'We have to guess it, but it\'s easy because most implementations of IPFS encode CIDs with "base32".',
```

src/tutorials/0006/05.vue
```diff
@@ -0,0 +1,42 @@
...
+    correct: false,
+    feedback: 'That doesn\'t sound right. Please review this lesson again.'
+  },
+  {
+    answer: 'We can check the initial characters of the hash. If it starts with "Qm", it\'s "base58btc". If it starts with "b" it\'s "base32". In both cases the CID version is "CIDv1"',
+    correct: false,
+    feedback: 'Hmm, not always. Please review this lesson again.'
+  },
+  {
+    answer: 'We can check the initial characters of the hash. If it starts with "Qm", it\'s "base58btc" and is "CIDv0." If it starts with "b" it\'s "base32" and is "CIDv1."',
```
alanshaw: Not quite correct. If we have a string and we think it's a CID then we check the first character to see if it's in the mutibase table. If it's in there then we're good, provided decoding succeeds. If it's not in the table then we just _try_ to decode it using `base58btc` because that's the only other thing it could be. So there's no checking of "Qm".

> If it starts with "Qm", it\'s "base58btc"

If it starts with "Qm" it's _not_ multibase encoded, but probably a multihash encoded using base58btc.

If it starts with "z" it's probably a base58btc multibase encoded CID.

Obviously it would be awkward for us if we ever defined an encoding in multibase of "Q".

src/tutorials/0006/06.md
```diff
@@ -0,0 +1,46 @@
...
+Since Version 0 CIDs did not have those prefixes, they are always assumed to be `base58btc` and `dag-pb` respectively.
+
+You may have noticed that the `Base32 CIDV1` part in the end is `bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi`, which is the same CID from the first example!
+
+Notice also how the end of the "Human Readable CID" (the portion after the prefixes) is exactly the same in this CIDv0 example as it was in the CIDv1 example: `c3c4733ec8affd06cf9e9ff50ffc6bcd2ec85a6170004bb709669c31de94391a`.
+
+Why? These two CIDs point to the same content. Basically, it's the same hash (`c3c4733ec8affd06cf9e9ff50ffc6bcd2ec85a6170004bb709669c31de94391a`) represented in the two different versions of the CID spec.
+
+You can convert any `CIDv0` to `CIDv1`, because the implicit prefixes from `v0` become explicit in `v1`.
+However, because `CIDv1` supports multiple codecs and multiple bases and `CIDv0` does not, not all `CIDv1` hashes can be converted to `CIDv0`. In fact, only `CIDv1` that have `multibase=base58btc` and `multicodec=dag-pb` can be converted to `CIDv0`.
```
alanshaw: ...and hash algo sha2-256 and hash length 256

https://github.com/multiformats/js-cid/blob/9fee0105e300c8e7cb9d14cda6823b62d61982eb/src/index.js#L192-L208

src/tutorials/0006/01.vue
```diff
@@ -0,0 +1,40 @@
...
+    :text="text"
+    :question="question"
+    :choices="choices" />
+</template>
+
+<script>
+import MultipleChoiceLesson from '../../components/MultipleChoiceLesson'
+import text from './01.md'
+
+const question = 'By default, what is the cryptographic algorithm used by IPFS to generate CIDs?'
```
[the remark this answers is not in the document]
terichadbourne: Can we come up with a phrase that doesn't use `default` twice but still makes @alanshaw notice that we used `default`? :)

src/tutorials/0006/04.md
terichadbourne: I can't believe you don't want to invent new versions of the CID! 😂

src/tutorials/0006/05.vue
terichadbourne: I believe this is a British v American English issue. If we can get markdown working here so we can use `base32` and put the period after it'll be perfect. :)

src/tutorials/0006/03.vue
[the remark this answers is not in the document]
zebateira: Mentioned in the lesson:
```
Usually, applications solve this issue by trying to guess the encoding based on its content, but this is not a deterministic or safe way to do it.
```

Multicodec tries to solve this problem as well with multicodec right?

src/tutorials/0006/01.md
```diff
@@ -0,0 +1,20 @@
+When we exchange data with peers on the decentralized web, we depend on **content addressing** (as opposed to the **location addressing** of the centralized web) to securely locate and identify data. If you haven't yet done so, please check out our [Decentralized Data Structures](https://proto.school/#/data-structures/) tutorial to learn about the basics of important decentralized web concepts like content addressing, cryptographic hashing, content identifiers (CIDs), and sharing with peers. In this tutorial we'll dig deeper into how CIDs are constructed in IPFS.
+
+A content identifier, or **CID**, is a hash used to point to material in [IPFS](https://ipfs.io). It doesn't indicate where the content is stored, but it forms a kind of address based on the content itself. CIDs are short, regardless of the size of their underlying content.
```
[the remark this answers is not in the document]
zebateira: sounds good. will revise this 👍

src/tutorials/0006/01.vue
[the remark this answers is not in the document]
zebateira: `What is the default cryptographic algorithm used by IPFS to generate CIDs?` sounds good?

src/tutorials/0006/02.md
```diff
@@ -0,0 +1,35 @@
+By using a cryptographic algorithm to generate the hash, we guarantee that we can identify a file with a single hash.
```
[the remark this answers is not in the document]
zebateira: Sounds good. Will remove this.

src/tutorials/0006/03.md
zebateira: I was under the impression that the multicodec content type on the CID could be _any_ multicodec id. That is why I didn't mention IPLD formats much. So you are saying it can only be an IPLD format id?

src/tutorials/0006/03.vue
zebateira: True, in theory might be possible. Will change this

src/tutorials/0006/01.vue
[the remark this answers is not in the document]
alanshaw: Oh, haha, my bad.

src/tutorials/0006/03.md
alanshaw: I guess techincally you could put _any_ multicodec there, but if you want to use the CID with IPFS then it _has_ to be a multicodec for an IPLD format.

src/tutorials/0006/05.vue
alanshaw: No way, I did not know this!

src/tutorials/0006/03.md
zebateira: I understand that in the context of IPFS, IPLDs are used, but since multiformats is project on its own, should we keep this generic just like the CID specification? Or should we contextualize this for IPFS only?

alanshaw: The purpose of IPLD is to have interoperable datasets. If you don't use an IPLD format codec then you're doing your own thing and that's not really the future we want to promote.

@terichadbourne should have the final say on this.

src/tutorials/0006/05.vue
zebateira: I see. I followed the documentation [https://github.com/multiformats/cid#decoding-algorithm](here) for this bit.
I also checked the implementations, and it seems that [go-cid](https://github.com/ipfs/go-cid/blob/master/cid.go#L236) and [rust-cid](https://github.com/multiformats/rust-cid/blob/master/src/version.rs#L23) follow the [decoding algorithm spec](https://github.com/multiformats/cid#decoding-algorithm) and that [js-cid](https://github.com/multiformats/js-cid/blob/master/src/index.js#L76) and [py-cid](https://github.com/ipld/py-cid/blob/master/cid/cid.py#L266) follow the algorithm that you described if I'm not mistaken.

So which one should we consider the correct one?

alanshaw: The spec - good detective work!

alanshaw: I opened an issue to get this fixed up https://github.com/multiformats/js-cid/issues/103

src/tutorials/0006/02.md
zebateira: @terichadbourne how do you think we could go about this?

src/tutorials/0006/01.md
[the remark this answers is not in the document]
terichadbourne: I'm not suggesting saying this in the lesson, just asking for myself @alanshaw, but is the part that's identical between v0 and v1 after you strip out all the prefixes actually a hash?

src/tutorials/0006/03.md
terichadbourne: Thanks for the clarification here, @alanshaw. Sounds like in the context of this tutorial focused on IPFS it would be useful to state that there's a generic thing called a multicodec and that in the case of IPFS we must always use IPLD format as our multicodec. Then for future references list `ipld-format` in the spots Alan has noted below. 

However, worth flagging here that the reason @zebateira referred to it as a multicodec is that he found the multicodec documentation in the multiformats project where CID now resides, and that documentation doesn't mention IPLD format at all. Is that an intentional documentation decision @mikeal (if I'm right that you're off-the-books-kinda-sorta-multiformats-guy in addition to IPLD guy? Is the IPLD format multicodec described somewhere else instead? Could we link between the two places in the docs?

FWIW I find it sooooooo confusing to be a learner in this ecosphere popping around in different repos of related projects. I think @zebateira  actually found the issue in which CID got moved from IPLD to Multiformats in his quest to find the right docs, so he deserves a detective magnifying glass trophy.

terichadbourne:
Just chatted with @alanshaw  about this grammatical concern of mine... 

> Throughout the tutorial, it feels odd to me when there are references to "which encoding was used" instead of "which encoding method" or "which type of encoding." I don't know if that's a genuine concern or is a matter of my technical vocabulary not being up to snuff. I haven't edited these spots because I'm not confident enough to do it, but maybe @alanshaw can weigh in when he reviews this.

He said that for someone technical it is totally fine to refer to "the encoding" as @zebateira did, and also totally fine to use my suggestions of "the encoding method" or "the type of encoding," but since we want to air on the side of accommodating less technical learners, we may as well go with my suggestions that will appear more grammatically correct to the newcomer. :)

src/tutorials/0006/03.vue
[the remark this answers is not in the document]
terichadbourne: I just chatted with @alanshaw about this distinction @zebateira  has been making between the trustworthiness of a CID versus the trustworthiness of magic numbers in normal encoded data/files used to guess at the file types. It sounds like he takes your point, and that it's true that if you always use CIDs in IPFS, you'll always know how to decode them without fumbling around. However, he thinks that referencing the way applications determine file types will make people think that this is about a different sort of encoding than it's about within IPFS. (We're not discovering that it's a PDF, we're discovering that it's dag-pb). So his recommendation is to remove the comparison to avoid people confusing the two processes with each other. (Feel free to correct me if I'm misrepresenting you, Alan.)

terichadbourne:
@zebateira I got `marked` working for the mutliple choice options, but it's actually adding a `<p>` wrapper around the contents (inside the `<label>`) and therefore adds default top/bottom margin to the paragraph. I can't seem to select it with `label > p` to remove the margin. Do you have suggestions for either fixing the CSS or using `marked` options to strip out the `<p>` tags?

image

I also edited the answers to use backticks where it seemed helpful.

src/tutorials/0006/03.vue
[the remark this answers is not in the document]
zebateira: sounds good 👌

zebateira:
@terichadbourne fixed 👍 
I simply removed them from the final string. It's not the ideal solution, but it works.

src/tutorials/0006/03.vue
zebateira: Changed to `The codec used on CIDv1 is always "base58btc".`

src/tutorials/0006/01.vue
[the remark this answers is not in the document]
zebateira: changed to: `What is the default cryptographic algorithm used by IPFS to generate CIDs?`

src/tutorials/0006/03.md
zebateira: added at the end of the lesson: `To note that, within IPFS, only IPLD formats are used so that means the multicodec prefix will always be an IPLD codec.`

zebateira:
Added the changes regarding the IPLD format mentions here: https://github.com/ProtoSchool/protoschool.github.io/compare/content/new-tutorial-anatomy-of-a-cid...content/new-tutorial-anatomy-of-a-cid-ipld?expand=1

terichadbourne:
We seem to have a bug that occurs if you get the correct answer and then get an incorrect answer. 

The correct state for a wrong answer is that the next button is disabled and the "try again" progress indicator is displayed. However, if you first pass the lesson, the `lessonPassed` prop seems not to be updating afterward in any of the components, so you see an enabled button and "you did it!" until you refresh the page. I checked under the hood and `clearPassed` is running and localStorage is working as it's supposed to, so it's just the Vue data that's not reacting. 

I tried turning lessonPassed into a computed property but it made the whole Quiz component disappear again. 

I don't recall seeing this error the first time I reviewed the PR, so not sure if it was introduced when dealing with merge conflicts or if I just didn't check the wrong-after-right situtation. 


I'm also getting a lot of errors and warnings in the console on this branch that are probably caused by earlier restructures to the codebase. An example of me loading a multiple choice lesson: 
image

terichadbourne:
@zebateira I took another full read through this one as an end user, fixed a few little grammatical things, updated feedback to give more clues, and added a little more content around the IPLD format multicodecs, etc. Let me know if anything doesn't sound good to you. I'd like to get one or two more volunteers to read through this early next week for some fresh eyes before we publish, and we have some bugs to fix in the multiple choice functionality as noted above.

src/tutorials/0006-anatomy-of-a-cid/06.md
```diff
@@ -1,5 +1,5 @@
 ---
-    title: "Examples"
+    title: "One hash, multiple CID versions"
```
zebateira: I wouldn't rename this lesson as such because this should not be the focus of the lesson. The focus of the lesson is to give examples of CIDs.
The last note about the CIDs representing the same content is just a small detail.

Does it look like this is the focus of the lesson? Because if that's the case I think we should re-word it then.
But if you think having a "Examples" lesson is less important than this we can keep it as is.

zebateira:
@terichadbourne the multiple choice error you mentioned is fixed and the warnings have been suppressed for multiple choice options.

zebateira:
> I took another full read through this one as an end user, fixed a few little grammatical things, updated feedback to give more clues, and added a little more content around the IPLD format multicodecs, etc. Let me know if anything doesn't sound good to you. I'd like to get one or two more volunteers to read through this early next week for some fresh eyes before we publish, and we have some bugs to fix in the multiple choice functionality as noted above.

Awesome 👏  looking good.

src/tutorials/0006-anatomy-of-a-cid/06.md
terichadbourne: The things I find most interesting about this lesson are: 
- the introduction of the CID Inspector tool 
- what I learn about the differences between CIDv0 and CIDv1 and how they both refer to the same underlying hash 

If you had named this Examples of _____ instead of Examples, what would you have named it? 

I don't find "Examples" alone to be a very compelling lesson title. It could be "Putting it all together" or "CID Inspector" or something about comparing CIDv0 to CIDv1. What do you think?

terichadbourne:
> @terichadbourne the multiple choice error you mentioned is fixed and the warnings have been suppressed for multiple choice options.

I don't see a commit for this fix. Has it been pushed?

src/tutorials/0006/05.md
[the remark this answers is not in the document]
ribasushi: @zebateira actually it's a bit more involved than that. Consider the following set of examples ( the `b` is missing due to cyberchef not being awesome ):
1. [DAG-PB through sha256](https://gchq.github.io/CyberChef/#recipe=From_Hex('Auto')To_Base32('a-z2-7%3D')&input=XDAxIFw3MCBcMTIgXDIwIFxmZg) ( currently most prevalent )
2. [raw leaves through sha256](https://gchq.github.io/CyberChef/#recipe=From_Hex('Auto')To_Base32('a-z2-7%3D')&input=XDAxIFw1NSBcMTIgXDIwIFxmZg) ( will be default for all leaf nodes in ipfs 0.6 )
3. [DAG-CBOR through sha256](https://gchq.github.io/CyberChef/#recipe=From_Hex('Auto')To_Base32('a-z2-7%3D')&input=XDAxIFw3MSBcMTIgXDIwIFxmZg) ( definitely what we will switch to wholesale before ipfs 1.0 )

Same as above but with blake2b ( likely candidate for ipfs 1.0 )
4. [DAG-PB/blake](https://gchq.github.io/CyberChef/#recipe=From_Hex('Auto')To_Base32('a-z2-7%3D')&input=XDAxIFw3MCBcYTAgXGU0IFwwMiBcMjAgXGZm)
5. [raw/blake](https://gchq.github.io/CyberChef/#recipe=From_Hex('Auto')To_Base32('a-z2-7%3D')&input=XDAxIFw1NSBcYTAgXGU0IFwwMiBcMjAgXGZm)
6. [DAG-CBOR/blake](https://gchq.github.io/CyberChef/#recipe=From_Hex('Auto')To_Base32('a-z2-7%3D')&input=XDAxIFw3MSBcYTAgXGU0IFwwMiBcMjAgXGZm)

Then there is *inlining/identity-hashing* which has yet another set of worms. The only thing you can *somewhat* hint at is `baf` for "maybe a CIDv1", due to the fact that the most widely used codecs are `0x55`, `0x70` and `0x71` ( `bafk...`, `bafy...`, `bafy...` respectively )

But at the end of the day @terichadbourne is correct: you should really only emphasize `b` and that's that

src/tutorials/0006-anatomy-of-a-cid/05.md
```diff
@@ -0,0 +1,32 @@
...
+
+[Multibase](https://github.com/multiformats/multibase) prefixes, which represent the base encoding used when converting CIDs between string and binary formats, are **only used in the string form of the CID**:
+
+![Multibase Prefix](tutorial-assets/T0006L05-multibase-prefix.jpg)
+
+Let's examine two examples of CIDs in their string form:
+
+![String form examples](tutorial-assets/T0006L05-string-form.jpg)
+
+We know the first one is a `CIDv0` because it starts with `Qm...`. All hashes that start with `Qm` can safely be interpreted as `base58btc` as a CID of Version 0.
```
ribasushi:
```suggestion
We know the first one is a `CIDv0` because it starts with `Qm...`. All CIDs that start with `Qm` can safely be interpreted as `base58btc` as a CID of Version 0.
```

src/tutorials/0006-anatomy-of-a-cid/06.md
```diff
@@ -0,0 +1,56 @@
...
+
+Why? These two CIDs point to the same content. Basically, it's the same hash (`c3c4733ec8affd06cf9e9ff50ffc6bcd2ec85a6170004bb709669c31de94391a`) represented in the two different versions of the CID spec.
+
+You can convert any `CIDv0` to `CIDv1`, because the implicit prefixes from `v0` become explicit in `v1`.
+However, because `CIDv1` supports multiple codecs and multiple bases and `CIDv0` does not, not all `CIDv1` can be converted to `CIDv0`. In fact, only `CIDv1` that have the following properties can be converted to `CIDv0`:
+
+- `multibase = base58btc`
+- `multicodec = dag-pb`
+- `multihash-algorithm = sha2-256`
+- `multihash-length = 256`
```
ribasushi: Probably worth doing an explicit call out here
```suggestion
- `multihash-length = 32 ( 256 bits )`
```

src/tutorials/0006-anatomy-of-a-cid/02.md
```diff
@@ -0,0 +1,36 @@
...
+
+A [**multihash**](https://multiformats.io/multihash/) is a self-describing hash which itself contains metadata that describes both its length and what cryptographic algorithm generated it. IPFS is future-proof because it uses multihash to
+**support multiple hashing algorithms** rather than relying on a specific one.
+
+Multihashes follow the `TLV` pattern (`type-length-value`). Essentially, the "original hash" is prefixed with a `type` and the `length` of the hash.
+
+![Multihash format](tutorial-assets/T0006L02-multihash.jpg)
+
+ - `type`: identifier of the **cryptographic algorithm** used to generate the hash (e.g. the identifier of `sha2-256` would be `18` - `0x12` in hexadecimal) - see the [multicodec table](https://github.com/multiformats/multicodec/blob/master/table.csv) for all the identifiers
+ - `length`: the actual **length** of the hash (using `sha2-256` it would be `256`)
```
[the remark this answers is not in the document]
terichadbourne: @olizilla Could you shed some light on why the  CID Inspector expresses length in `bits` vs `bytes`? @ribasushi is suggesting switching to `bytes` but we're explaining the CID Inspector readout here so we'd confuse the reader by showing something different from what's listed there.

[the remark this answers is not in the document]
ribasushi: @terichadbourne note that I am not specifically advocating for switching from bits to bytes in the *visualiser*. It is the on-wire format where everything is unquiestinably in bytes ( this also has an implication that you can not take 255 bits of the hash: you can do either 248 or 256 and that's that )

src/tutorials/0006-anatomy-of-a-cid/06.md
zebateira: Ah I see. That makes sense. I agree with `One hash, multiple CID versions` then!

src/tutorials/0006-anatomy-of-a-cid/01.md
```diff
@@ -0,0 +1,25 @@
+---
+    title: "What is a CID?"
+    type: "multiple-choice"
+---
+
+When we exchange data with peers on the decentralized web, we depend on **content addressing** (as opposed to the **location addressing** of the centralized web) to securely locate and identify data. If you haven't yet done so, please check out our [Decentralized Data Structures](https://proto.school/#/data-structures/) tutorial to learn about the basics of important decentralized web concepts like content addressing, cryptographic hashing, content identifiers (CIDs), and sharing with peers. In this tutorial we'll dig deeper into how CIDs are constructed in IPFS.
+
+A content identifier, or **CID**, is a hash used to address data stored on [IPFS](https://ipfs.io). It doesn't indicate where the content is stored, but it forms a kind of address based on the content itself. CIDs are short, regardless of the size of their underlying content.
```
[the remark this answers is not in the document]
zebateira: Agree!

src/tutorials/0006/05.md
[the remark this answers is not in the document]
zebateira: @ribasushi makes sense 👍 will update then!

src/tutorials/0006-anatomy-of-a-cid/05.md
```diff
@@ -0,0 +1,32 @@
...
+
+![Multibase Prefix](tutorial-assets/T0006L05-multibase-prefix.jpg)
+
+Let's examine two examples of CIDs in their string form:
+
+![String form examples](tutorial-assets/T0006L05-string-form.jpg)
+
+We know the first one is a `CIDv0` because it starts with `Qm...`. All CIDs that start with `Qm` can safely be interpreted as `base58btc` as a CID of Version 0.
+
+The second one starts with `bafy...`. The first character `b` is the base encoding prefix indentifier for `base32`, which is used by default by most implementations of IPFS.
```
zebateira:
```suggestion
The second one starts with `b...`. The character `b` is the base encoding prefix identifier for `base32`, which is used by default by most implementations of IPFS.
```

zebateira: @terichadbourne @ribasushi

zebateira:
> > @terichadbourne the multiple choice error you mentioned is fixed and the warnings have been suppressed for multiple choice options.
> 
> I don't see a commit for this fix. Has it been pushed?

Yeah: https://github.com/ProtoSchool/protoschool.github.io/pull/359/commits/f42114ec66753063a82b7889555189231e8c4ec6

src/tutorials/0006-anatomy-of-a-cid/05.md
ribasushi: Yup, this is it! Fixing the images would be great too, but if too much work: meh 😺

src/tutorials/0006-anatomy-of-a-cid/06.md
```diff
@@ -0,0 +1,56 @@
...
+
+To test this theory, you can check out our beloved `Aardvark` page here, hosted on the IPFS network: https://ipfs.io/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/wiki/Aardvark.html
+
+- Open the link in your browser and copy the CID from the middle of the URL (`QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco`)
+- In a new browser window, paste it into the [CID Inspector tool](https://cid.ipfs.io) and find the equivalent CIDv1 value displayed at the bottom of the screen
+- Back in your aardvark tab, replace the `v0` CID with the converted `v1` CID in the original URL and refresh the page
+
+You should see the same article on our aardvark.
```
johnnymatthews: I'd like to see a summary here explaining the final anatomy of a CID with an illustration. Something like this:

```markdown
Summary

So when presented with a CID, you should now be able to find the following information:

1. The hash of the content itself.
2. The length of that hash.
3. The hashing function used to hash the content.
4. The encoding method used to read the content.
5. The CID version this CID is built from.
6. And if the CID is written as a string (not as binary), what type of base encoding was applied to the binary data.

![Example of a CID with each section listed and coloured.]()
```

The image could be something like this hastily draw picture courtesy of Microsoft paint:

Untitled

This image is 100% not accurate. I just threw it together to illustrate what I think would work best here.

johnnymatthews:
I've submitted my suggestions as comments for some reason, but they're there for you to take a look at. Overall this is an incredible tutorial, you've done a great job here ProtoSchool!

src/tutorials/0006-anatomy-of-a-cid/06.md
ribasushi: @johnnymatthews sadly this is not possible: there is no direct character-for-character correlation between a base32 encoding and the underlying bytes.

Perhaps we could set up a 15 min demo with cyberchef to give you folks some intuition on how the pieces build on top of each other.

src/tutorials/0006-anatomy-of-a-cid/03.md
```diff
@@ -0,0 +1,24 @@
...
+But when we're trying to read the data itself, how do we know the encoding method used?
+It could be encoded with CBOR, Protobuf, plain JSON, etc. Usually, applications solve this issue by trying to guess the encoding based on its content, but this is not a deterministic or safe way to do it.
+
+To solve this issue, `CIDv1` introduces another prefix that uniquely identifies the encoding method used.
+
+## Multicodec prefix
+
+The **multicodec prefix** indicates which encoding was used on the data.
+
+![Multicodec Prefix](tutorial-assets/T0006L03-multicodec.jpg)
```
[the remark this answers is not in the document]
ribasushi: @johnnymatthews the 2-byte encoding is a result of the 256bit/256byte mistake I highlighted earlier. 256 encoded as a varint is [`128 2`](https://gchq.github.io/CyberChef/#recipe=VarInt_Encode()To_Binary('Space')&input=MjU2)

src/tutorials/0006-anatomy-of-a-cid/01.md
```diff
@@ -0,0 +1,25 @@
+---
+    title: "What is a CID?"
+    type: "multiple-choice"
+---
+
+When we exchange data with peers on the decentralized web, we depend on **content addressing** (as opposed to the **location addressing** of the centralized web) to securely locate and identify data. If you haven't yet done so, please check out our [Decentralized Data Structures](https://proto.school/#/data-structures/) tutorial to learn about the basics of important decentralized web concepts like content addressing, cryptographic hashing, content identifiers (CIDs), and sharing with peers. In this tutorial we'll dig deeper into how CIDs are constructed in IPFS.
```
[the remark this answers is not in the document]
terichadbourne: This tutorial is about the **anatomy** of a CID - super deep dive into all those dang prefixes. The referenced lesson covers more topics (as mentioned) including the broader context of content addressing. I don't want to re-teach all of content v location addressing here but I think it's super important for context.

terichadbourne:
Flagging two suggestions from a Slack convo w/ @mikeal & @ribasushi today: 
-  change `<ipld-format>` to `<ipld-codec>` when spelling out the various prefixes (@ribasushi has already suggested some in-line edits)
- change this whole tutorial from being labeled as IPFS to being labeled as Multiformats (we’ll need to add the project logo, fix the project metadata, and rework some of the text )

I haven't had time to go through all of @ribasushi & @johnnymatthews edits yet but really appreciate all the feedback!

src/tutorials/0006-anatomy-of-a-cid/06.md
johnnymatthews: Ah, that's a shame. Would the above be possible for a binary implementation?

src/tutorials/0006-anatomy-of-a-cid/01.md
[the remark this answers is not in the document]
johnnymatthews: Ok cool, that makes sense.

src/tutorials/0006-anatomy-of-a-cid/06.md
ribasushi: Yep! ( if by `implementation` you meant `representation` )

Basically for what you are asking to work, the requirement is that **the source base** ( in our case `256`: 8 bits in a byte == `2^8` == `256` ) **can be reduced by using only the √** operation to **the target base**. Therefore in your case all you have is`16`, `4`, and `2`: i.e. hex( `\x01` `\xE2` ... ) , nibbles ( `\x1` `\xA` `\xF` ... ) and binary ( `1` `1` `0` `1` ... ). Any other base ( 32, 58, etc ) will be "misaligned" by a certain cyclic amount of characters.

This is also why you often hear that our "native" base of 10 doesn't fit very well into computers ( √10 is not an integer )

ribasushi: @johnnymatthews another way to look at it for bytes vs base32 encoding: https://upload.wikimedia.org/wikipedia/commons/f/fe/Base32_DE.svg

zebateira:
> change this whole tutorial from being labeled as IPFS to being labeled as Multiformats (we’ll need to add the project logo, fix the project metadata, and rework some of the text )

@terichadbourne added the support for multiple projects in #396 which should address this

src/tutorials/0006-anatomy-of-a-cid/02.md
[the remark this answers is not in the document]
olizilla: I made the CID inspector show bits as it was heavily inspired by https://github.com/multiformats/multihash#idea-self-describe-the-values-to-distinguish

I don't have a strong opinion on if the length should be show in `bits` or `bytes` as long as it includes the unit. bits seems popular when describing the length of hash functions https://en.wikipedia.org/wiki/List_of_hash_functions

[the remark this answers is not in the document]
terichadbourne: The particular lesson in the tutorial is about showing the CID Inspector and explaining what you'll find there, so I'd love to match what's there as opposed to introducing concepts that aren't there. 

To @olizilla's point, it's much easier for me to understand sha-256 matching something expressed as 256 X than as 32 Y.  "Normal people" don't know how to convert bits to bytes in their head so they won't connect 32 and 256 on their own. 

@ribasushi perhaps you could put in a PR to the CID Inspector to add a bytes field _in addition to_ the bits field to appease both the technical purists and the newcomers? And once it's there we could update the tutorial content to mention both?

src/tutorials/0006/03.md
mikeal: Let me try to simplify this a little bit.

`multicodec` is just a table of identifiers. There isn’t an enforced classification between the identifiers (informally we have some comment and ranges in the table), some are hash algorithms, some are serializable formats, and when you get into encryption they are potentially both.

A CID is just a multicodec prefix for the **codec** plus a multihash (which has another multicodec prefix for the hash algorithm), there isn’t much validation here for which part of the table the prefix is pulled from. You could create a CID with a multicodec prefix for a hash algorithm and it won’t end up failing until you try to lookup the codec implementation.

I wouldn’t use the term “IPLD format” because that’s also rather confusing. IPLD is about building data structures out links to any data you can address with a CID and that includes codecs that really have nothing to do with IPLD (JSON, git, etc). Even the term “IPLD codec” implies that these are built by/for IPLD when in many cases they are not.

I would just use the term **“codec.”** The main difference between the identifier used in the CID and the one used for multihash is that the data being addressed by the CID should be parsable by a codec for the given the identifier. That’s it. There’s nothing IPLD specific about CID’s and their codec identifiers.

Being that IPLD isn’t a single project but an umbrella for things we do with content addressed data structures, it’s often hard to figure out where the lines are and when something should be called “IPLD.” CID is right on the edge because it’s what we use to link data together and once you’re liking data together, that’s IPLD. However, you *could* use CID’s **without linking anything together**. You could have a database of hashed JSON blocks and all the id’s are CID’s with a JSON multicodec prefix. Perfectly valid use of CID, but nothing is ever linked together, so it’s not really “IPLD.”

src/tutorials/0006-anatomy-of-a-cid/02.md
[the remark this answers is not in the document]
ribasushi: @terichadbourne I think we miscommunicated somewhere. I believe the CID Inspector **is doing the right thing** - i.e. speaking of bits in the context of hashes. In other words: what @olizilla said.

What I was highlighting is that **on the wire** you have to work in bytes. Since a number of your images show the on-wire binary, you need to:
- fix the images to show 0x20
- (optionally) explain why 0x20 and 256 are the same thing in this context

TLDR:
- when visualizing: speak in bits
- when talking about "what is *in* the CID" - you need to show/mention bytes ( otherwise https://github.com/ProtoSchool/protoschool.github.io/pull/359#discussion_r384689415 happens )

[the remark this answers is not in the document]
terichadbourne: @ribasushi To clarify, is it that there are pictures of 1s and 0s that have the wrong order of 1s and 0s in them because some conversion was done incorrectly or that you believe certain parts of the  1s and 0s pics should be labelled as "0x20" instead of labelled as "256"?

terichadbourne:
Quick update after a chat with @ribasushi: 

His suggestion is that in this image, the `128` should say `32` and that the whole colored section labelled as `2` should be gone (the beginning of the hash itself is what would come next):

image

ribasushi:
> His suggestion is that in this image, the `128` should say `32` and that the whole colored section labelled as `2` should be gone (the beginning of the hash itself is what would come next):
> 
> image

And ideally the bits should also express 32, or [`00100000`](https://gchq.github.io/CyberChef/#recipe=VarInt_Encode()To_Binary('Space')&input=MzI).

I want to stress again that I understand that the schedule is tight, and there is a number of images to fix. As long as the group makes an informed decision to leave things as-is, having understood the mismatch: I am fine with this 😸

alanshaw:
> > His suggestion is that in this image, the `128` should say `32` and that the whole colored section labelled as `2` should be gone (the beginning of the hash itself is what would come next):
> > image
> 
> And ideally the bits should also express 32, or [`00100000`](https://gchq.github.io/CyberChef/#recipe=VarInt_Encode()To_Binary('Space')&input=MzI).
> 
> I want to stress again that I understand that the schedule is tight, and there is a number of images to fix. As long as the group makes an informed decision to leave things as-is, having understood the mismatch: I am fine with this 😸

This is my mistake. It should really be fixed since it's the "anatomy of a CID" tutorial. The length should be 32 (`00100000` in binary varint encoded form) instead of 128 and 2, as verified here:

Screenshot 2020-02-27 at 23 02 00

(Note the first number, 18 is [sha2-256](https://github.com/multiformats/js-multicodec/blob/8815d4b8087fb24e15fa6bb4e589316aac45286b/src/base-table.json#L6))

alanshaw:
FWIW **IPLD formats** are mentioned quite a bit [in](https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-pb.md#format) [the](https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-cbor.md#format) [specs](https://github.com/ipld/specs/blob/master/block-layer/serialization-and-formats.md) and extensively in the current [js-ipld API](https://github.com/ipld/js-ipld#table-of-contents) used by js-ipfs.

If these are not up to date and it's not a term being used anymore then please forgive me but I knew no better 🙏

src/tutorials/0006-anatomy-of-a-cid/02.md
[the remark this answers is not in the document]
zebateira: Very good point @ribasushi.
@alanshaw mentioned in the original talk that the representation is like so (1st byte = 128, 2nd byte = 2) because of the varint representation. But I thought it was not necessary to bring that up in this tutorial.

So regarding the length mentions, we should change it to bytes (show the conversion from bits to bytes like you said) - 100% agree.
Regarding the binary representation, do you think it would make sense to put the two bytes together and point to `256` then? Maybe with `varint` under parentheses?

zebateira:
To sum up everything of the feedback:

 - [ ] Updating visuals and content: representation of the hash in binary should be changed to represent `32 bytes (256bits)` and we should mention the units always as well: either bytes and/or bits.
 - [ ] Updating `bafy...` highlight in the images to `b...`
 - [ ] Decide on whether we should mention the multicodec prefix as `ipld-format`, `ipld-codec` or `codec`.

As mentioned by @ribasushi these edits can be fixed after we publish, which might not be a bad idea considering the length of this review and timeline for publishing this tutorial.

src/tutorials/0006-anatomy-of-a-cid/02.md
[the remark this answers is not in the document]
ribasushi: We had a short Zoom with @zebateira to clarify the context of above remark. Desired outcome is [mention the appropriate/expected unit in the corresponding context](https://github.com/ProtoSchool/protoschool.github.io/pull/359#issuecomment-592529505).

Thanks for taking the time to understand the subtle mismatch!

terichadbourne:
The other missing piece in your list is some light reframing to fit the new Multiformats project label. I can take that on while you're out @zebateira.

@alanshaw are the "images" in your preso actually image files or are they Google Slides text boxes and arrows and such? Could I snag a copy of the original Google Slides file? Ze has been working from the PDF you shared in the camp repo.

terichadbourne:
Latest commit updates images as follows based on recommendations from @ribasushi & @alanshaw. Please let me know if anything isn't as intended.

old
image

new
image

old
image
new
image

old
image

new
image

old
image

new
image

src/tutorials/0006-anatomy-of-a-cid/06.md
terichadbourne: @johnnymatthews Understanding that we couldn't do it to a picture of the string CID and would have to do it to a picture of the binary, would you still find this summary diagram helpful? I'm assume it would be some combination of these three formats of diagram, but extended out to include every type of prefix except base prefix, since that only exists in strings: 
image

Note that we already have 
image
and 
image
as our closest-to-complete diagrams in other formats throughout the lessons.

terichadbourne:
@mikeal I believe we've now made all the changes necessary to reframe this as a Multiformats tutorial. Would you have a sec to peruse as an end user (which will be much easier than reviewing the full code diff) and let me know if anything strikes you as incorrect or poorly explained?

ribasushi:
@terichadbourne all the images in https://github.com/ProtoSchool/protoschool.github.io/pull/359#issuecomment-593612290 are correct, thank you 🎉 
Note that the images at https://github.com/ProtoSchool/protoschool.github.io/pull/359#discussion_r386670140 have the old (incorrect) 256-based value though.

src/tutorials/0006-anatomy-of-a-cid/05.md
```diff
@@ -0,0 +1,32 @@
...
+
+`<cid-version><ipld-format><multihash>`
+
+Since binary CIDs are not very human-friendly, we can represent these binary CIDs in a string form (binary data represented as text). Example:
+
+`bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi`
+
+Converting data between binary and string formats requires base encoding, so when working with string CIDs it's important that we know the type of base encoding that was applied to the binary data. But how can we identify this?
+
+In `CIDv0`, hashes are **always** encoded with `base58btc`. _Always_. This means that we can safely interpret `CIDv0` hashes assuming they are using `base58btc`. But, for performance reasons, we may want to support multiple base encodings. For that, you guessed it, we can add another prefix!
```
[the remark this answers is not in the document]
ribasushi: @zebateira @terichadbourne ( FOR LATER!!! ) please consider incorporating this change into a future version of the doc. There is no practically noticeable performance impact between base58btc and base32 encoding: the main reason is usability.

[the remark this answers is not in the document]
terichadbourne: updated

terichadbourne:
Shipping it! Please feel free to open issues for any suggested improvements moving forward. 

Draft of the announcement blog post can be found here: https://github.com/ipfs/blog/pull/379

PR #360, opened 2020-01-28
Fix tutorial URLs

terichadbourne:
@zebateira A couple of folks used the new Twitter feature recently but I clicked through on the links and reached a 404 because the hash was missing in the URL. Feel free to merge if you agree that this little tweak fixes the issue. 

image

Interestingly hitting a GitHub-generated 404 and not our own - will this case be fixed by the routing PR you have in progress by any chance?

image

### DIFF
--- a/src/utils/tutorials.js
+++ b/src/utils/tutorials.js
@@ -5,6 +5,6 @@ export const getCurrentTutorial = route => (
     tutorialsList[route.props.default.tutorialId]
 )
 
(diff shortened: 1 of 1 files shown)


zebateira:
Ah nice catch!

> Interestingly hitting a GitHub-generated 404 and not our own - will this case be fixed by the routing PR you have in progress by any chance?

hm no, unfortunately, when you access any route besides `/` you will get github's 404.
but we can customize it using: https://help.github.com/en/github/working-with-github-pages/creating-a-custom-404-page-for-your-github-pages-site

PR #370, opened 2020-02-10
feat: increase horizontal padding in all sections from ph2 to ph3

zebateira:
This way the content will not be so close to the viewports width on smaller devices.
Also made other ui tweaks and fixed some alignments. @terichadbourne feel free to reject them.

# Homepage

## Before vs After
<div>

</div>

# Tutorial Page

## Before vs After

<div>

</div>


terichadbourne:
Lots of great visual tweaks here! 

One little accessibility concern: If I navigate using a keyboard in the new version, it's very hard to see the ring around the lesson I've selected, whereas in the old version it highlights yellow as it would if you hovered. Mind quickly updating to use your new color scheme both ways?

Old (hover and keyboard focus): 
image

New (keyboard focus): 
image

New (hover): 
image

zebateira:
Nice catch!
Fixed:
image

PR #375, opened 2020-02-14
fix: mark resources "lesson" as passed on page mount

### LINKED ISSUE #368: Bug: Resources page is only marked as completed if the user clicks on "More Tutorials" button at the bottom of the Resources Page
We should mark the tutorial as complete as soon the last lesson is completed successfully.

Mentioned in #359

zebateira:
This PR fixes the issue of when the resources lesson only gets marked as completed after you click the button "More Tutorials" at the end of the page.
For this we mark the lesson as completed when the page mounts

Closes #368

### DIFF
--- a/src/components/Lesson.vue
+++ b/src/components/Lesson.vue
@@ -265,6 +265,12 @@ export default {
   beforeMount: function () {
     this.choice = localStorage[this.cacheKey] || ''
   },
+  mounted: function () {
+    if (this.isResources) {
+      localStorage[this.lessonKey] = 'passed'
+      this.trackEvent(EVENTS.LESSON_PASSED)
+    }
+  },
   methods: {
     setEditorCode (newCode) {
       localStorage[this.cacheKey] = newCode
--- a/src/components/Lesson.vue
+++ b/src/components/Lesson.vue
@@ -469,7 +469,7 @@ export default {
         key: event,
         segmentation: {
           tutorial: this.tutorial.shortTitle,
-          lessonNumber: this.lessonId,
+          lessonNumber: this.isResources ? 'resources' : this.lessonId,
           path: this.$route.path,
           ...opts
         }


src/components/Lesson.vue
```diff
@@ -265,6 +265,13 @@ export default {
   beforeMount: function () {
     this.choice = localStorage[this.cacheKey] || ''
   },
+  mounted: function () {
+    if (this.isResources) {
+      localStorage[this.lessonKey] = 'passed'
+      this.trackEvent(EVENTS.LESSON_PASSED)
+      this.updateTutorialState()
```
terichadbourne: I believe this line isn't necessary because `updateTutorialState` only checks real numbered lessons, not the resources lesson, so completion of the resources page has no affect on tutorial completion, right?

terichadbourne:
The new code is working for me. 

However, just one point of clarification on issue #368. I believe that the tutorial is already marked as passed when the last REAL lesson is completed, based on the way the looping works in `setTutorialState`. So the bug we're fixing here is that the resources page is "**Resources lesson** is only marked as completed if the user clicks on "More Tutorials" button at the bottom of the Resources Page," not ": **Tutorial** is only marked as completed if the user clicks on "More Tutorials" button at the bottom of the Resources Page." Do you agree? 

From an analytics standpoint, people finishing all of the "real" content feels like the better indicator of whether a tutorial is completed, so I prefer it as is. 

Based on this I think there's one line of extraneous code being run on mount, as called out in a comment. 

Let me know if you're on the same page about how this is working.

src/components/Lesson.vue
zebateira: correct 👌

zebateira:
@terichadbourne you are correct. updated 👌

terichadbourne:
Thanks! I  think it'll be better to leave the `this.trackEvent(EVENTS.LESSON_PASSED)` for reporting on the resources page (I know there's no real distinction between this and a page view but it'll be easier to see the lessons passed stats in one place in reports) and only remove the `setTutorialState`. Does that make sense to you?

terichadbourne:
That way we can get resources to show here: 
image

zebateira:
@terichadbourne oh ok, sorry, misread what you meant. updated 👍

terichadbourne:
Made one more tweak to fix Countly. Previously it was catching resources pages in the `lessonPath` events by segment `path` but not by segment `lessonId`, so you couldn't see it alongside other lessons in certain dashboards. This change sends a lessonId of `resources` if `isResources` is true, otherwise a real `lessonId`. We can now see the lesson completion under both the `path` and `lessonId` segments, as tested in Countly.

zebateira:
@terichadbourne perfect 👌

PR #376, opened 2020-02-14
chore(docs): some boilerplate updates with the new options

zebateira:


terichadbourne:
@zebateira - I pushed a commit with some suggested edits to give a bit more context. Let me know what you think! 

There's currently no detailed documentation for the `modules` option, so I think we should add that before merging this. Some lessons, like in Basics, use an import statement to access CID, and a bunch of other lessons say they're using the CID module via the option but don't actually seem to use it.

zebateira:
@terichadbourne updated 👍

terichadbourne:
@zebateira I tweaked your formatting in a way that would make more sense to me as a reader - let me know if this still works.

src/tutorials/boilerplates/boilerplate-code.js
```diff
@@ -31,8 +31,8 @@ return run
 `
 
 // Optional: Use the modules option if you need extra
-// modules available in the solution code
-const modules = { cids: require('cids') }
+// modules available in the validation code
```
zebateira:
```suggestion
// modules available in the solution code
```

zebateira: @terichadbourne  Should be solution code, not validation. Basically modules available to the user when writing up the solution code.

```suggestion
// modules available in the solution code
```

zebateira:
@terichadbourne all good. just one last comment

PR #381, opened 2020-02-17
Upgrade to ipfs@0.41

### LINKED ISSUE #347: Update all tutorials to reflect async/await refactors to IPFS API
@alanshaw called out that the async/await changes being made to js-ipfs in https://github.com/ipfs/js-ipfs/pull/2683 will make our lesson content outdated. (Nothing on the site will actually break until we update IPFS versions.) I'll schedule some time for us to chat about the scope of the changes and who's best equipped to make updates.

zebateira:
Closes #347


zebateira:
@terichadbourne so I've created two PRs for us to see the differences between the two options that I see moving forward:

1. #382 Use `for await...of` loop

Screen Shot 2020-02-17 at 7 04 37 PM

- Still working out how to solve the linting warnings on the monaco editor
- Although it looks a bit convoluted, it might be a better choice if we don't want to introduce a module when teaching

2. #383 Use `it-all` package

Screen Shot 2020-02-17 at 7 06 01 PM

- A bit more clean but it might be confusing for users to understand where to use the `all` function

zebateira:
IMHO, I prefer the second option since it reduces the complexity of the code a lot. We need to consider that the last lessons of the tutorials would get much more complex with a lot of for loops if we don't use something like `it-all`.
We just need to explain to the user why we need to use it and include it in the initial code solution as much as possible.

zebateira:
### Update

Regarding all the points mentioned in https://github.com/ProtoSchool/protoschool.github.io/pull/385#issuecomment-587961261 I think I've addressed them all except for:

- More detailed explanation about `Async Iterables`
- Add small observation to users that have completed the tutorial in the past and are now coming back to it: the cached code will not work anymore

zebateira:
@terichadbourne ready for review, will address the "warning" message in another PR

terichadbourne:
@alanshaw Thanks for offering to give this one a proof! 

Note that we'll also be adding a few callouts that appear only to folks who already visited the previous version of the affected tutorials, so what you're seeing is the text for new visitors.

alanshaw:
Just a quick drive by comment...I'm hesitant to use `it-all` in examples or tutorials because of https://tc39.es/proposal-iterator-helpers/#sec-asynciteratorprototype-toarray - which will eventually allow you to `await ipfs.add(files).toArray()` to do the same thing (more info here https://github.com/tc39/proposal-iterator-helpers).

terichadbourne:
Any sense of how long "eventually" will be @alanshaw? I'd argue that if "using a thing that lets you nicely turn stuff into an array" is the preferred/easier method we want to teach, then rather than rip out the `it-all` and making the lessons harder by requiring all the JavaScript looping it might make more sense to add a note about the hopefully-upcoming-built-in-thing and let people know it will (hopefully) produce the same results as `it-all` does and we'll come back and update the content when that becomes available. 

Probably worth us both working through the content in full to see how it flows before we decide.

alanshaw:
> Any sense of how long "eventually" will be @alanshaw?

Not at all. It is currently at [stage 2](https://tc39.es/process-document/).

> I'd argue that if "using a thing that lets you nicely turn stuff into an array" is the preferred/easier method we want to teach, then rather than rip out the `it-all` and making the lessons harder by requiring all the JavaScript looping it might make more sense to add a note about the hopefully-upcoming-built-in-thing and let people know it will (hopefully) produce the same results as `it-all` does and we'll come back and update the content when that becomes available.

That's fair, and I have no desire to make lessons harder, but if we're going to be turning stuff into an array and then looping over it we should consider just using a for loop instead (for example). I'll take a look tomorrow but I expect @zebateira has made sensible choices already.

alanshaw:
...there's also the small problem of `require` not being available in the browser and all the explanation you'd need to do around that 😅

terichadbourne:
Little UI bug: There used to be a white box around the log output, which still exists on the live site but not in this PR (and IIRC not in the Anatomy of a CID PR): 
live site:
image
this branch:
image

terichadbourne:
@zebateira You had mentioned something about CID formatting in a chat recently. Just confirming the the format logged below is genuinely what's returned by IPFS, not you making it prettier or wrapping it in `CID()`? 
image

zebateira:
@alanshaw 

> ...there's also the small problem of require not being available in the browser and all the explanation you'd need to do around that 😅

Gosh I missed that. I guess we can make it global just like ipfs. @terichadbourne thoughts? Given this I think we should remove the `require` functionality and add the `all` to the globals as well.

zebateira:
> You had mentioned something about CID formatting in a chat recently. Just confirming the the format logged below is genuinely what's returned by IPFS, not you making it prettier or wrapping it in CID()?

@terichadbourne What I mentioned is that I _am_ formatting it to show `CID()` otherwise the output would be:

image

zebateira:
@terichadbourne 

> Little UI bug: There used to be a white box around the log output, which still exists on the live site but not in this PR (and IIRC not in the Anatomy of a CID PR):

Fixed!

zebateira:
@alanshaw 

> I'm hesitant to use `it-all`

1. Would be amazing to use `toArray()` but don't you think it should be at least stage 3 to make content in ProtoSchool using it? Unless it's a proposal that is popular in the community and has requested it a lot.

2. The memory issue: this note is in the first lesson that introduces `it-concat`

> When you're ready to try this in the real world, you should note that the above example can result in heavy memory usage, depending on the contents of the file being read. If you find this to be the case, you might want to skip the usage of the `it-concat` package and process each chunk of data iteratively.
> This is the main reason why ipfs returns `Async Iterables` so that we can decide if we want to be aware of memory performance issues or not.
> But in ProtoSchool tutorials, we don't really need to worry about this since our code challenges will use small files, so we can concatenate everything without worrying about performance.

src/tutorials/0004-mutable-file-system/05.md
```diff
@@ -23,20 +23,36 @@ such as `/catPics`,
...
 
 If we wanted to view all available details for the contents of a `/catPics`
 directory, we could do it like this:
 
 ```js
-await ipfs.files.ls('/catPics', { long: true })
+ipfs.files.ls('/catPics')
+```
+
+Because the `files.ls` method returns an [`Async Iterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of), you'll need to call the iterable multiple times until you have all the results.
```
alanshaw: Hmm, you don't really call it, unless you're interfacing with the raw `.next()` method - you "iterate over it's values". This says you _need_ to get all the results, which I guess is true for this exercise as it stands, but you can stop iterating when you like, you don't have to consume it all. For example, if you're looking for a particular thing you can stop early, and actually that's better for the IPFS node because it doesn't need to fetch everything (from disk and/or the network).

Instead of the exercise being to collect items in an array why not do one of:

1. Return the async iterable
2. Log out the name and size of each file with `console.log` in the `for await` loop. You could easily stub `console.log`, collect the results and assert on what was logged when the function completes.

I think the latter is more interesting, especially since earlier we have "If we wanted to view all available details for the contents of a `/catPics`" - currently we don't "view" anything.

src/tutorials/0004-mutable-file-system/05.md
```diff
@@ -23,20 +23,36 @@ such as `/catPics`,
...
+
+for await (const resultPart of ipfs.files.ls('/catPics')) {
+    result.push(resultPart)
+}
+```
+
+To make things easier, we can use the `it-all` package that does this automatically:
+
+```javascript
+const all = require('it-all')
```
alanshaw: If not already explained, this `require` magic in the browser needs some love.

src/tutorials/0004-mutable-file-system/10.md
```diff
@@ -8,23 +8,29 @@ MFS has a [`files.read`](https://github.com/ipfs/interface-js-ipfs-core/blob/mas
...
 
 or
 
 ```js
-let contents = (await ipfs.files.read('/directory/some-file.txt')).toString('utf8') // a string
+const concat = require('it-concat')
+
+let contents = (await concat(ipfs.files.read('/directory/some-file.txt'))).toString('utf8') // a string
 // notice the parentheses around the entire await statement
 ```
```
alanshaw: Personally I'd remove this second example - it's not super readable and doesn't help with much other than to save on 1 line of code.

src/tutorials/0004-mutable-file-system/10.md
```diff
@@ -8,23 +8,29 @@ MFS has a [`files.read`](https://github.com/ipfs/interface-js-ipfs-core/blob/mas
...
 
-When you're ready to try this in the real world, you should note that `files.read` method can result in heavy memory usage based on the contents of the file being read. You might want to explore the [`files.readReadableStream`](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/FILES.md#filesreadreadablestream) or [`files.readPullStream`](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/FILES.md#filesreadpullstream) methods instead.
+When you're ready to try this in the real world, you should note that the above example can result in heavy memory usage, depending on the contents of the file being read. If you find this to be the case, you might want to skip the usage of the `it-concat` package and process each chunk of data iteratively.
+This is the main reason why ipfs returns `Async Iterables` so that we can decide if we want to be aware of memory performance issues or not.
+But in ProtoSchool tutorials, we don't really need to worry about this since our code challenges will use small files, so we can concatenate everything without worrying about performance.
```
alanshaw:
```suggestion
In ProtoSchool tutorials, we don't really need to worry about this since our code challenges use small files, so we can concatenate everything without worrying about performance.
```

src/tutorials/0005-regular-files-api/03.md
```diff
@@ -19,33 +19,52 @@ When you add a file to IPFS, you're putting it only in your own node, but making
...
 
 Note that the `add` method can accept either a single `File` or an array, just in case you have more than one adorable animal photo to add to the node:
 
 ```javascript
-await ipfs.add([catPic, dogPic, giraffePic])
+ipfs.add([catPic, dogPic, giraffePic])
 ```
 
-Because the `add` method returns a `Promise`, you'll need to place an `await` before the method call to suspend the execution until the return value of the promise is ready to be used.
+Because the `add` method returns an [`Async Iterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of), you'll need to call the iterable multiple times until you have all the results.
```
alanshaw:
```suggestion
Because the `add` method returns an [`Async Iterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of), you'll need to collect all the results by iterating over it until it is done.
```

src/tutorials/0005-regular-files-api/03.md
```diff
@@ -19,33 +19,52 @@ When you add a file to IPFS, you're putting it only in your own node, but making
...
-    hash: string,
-    size: number
+    cid: CID,
+    size: number,
+    mode: number
 }
 ```
 
-The value of the `hash` is a CID (Content Identifier), a unique address generated from the content of the node. (For a more in-depth look at how CIDs are generated and why they're important, check out our [Decentralized data structures](https://proto.school/#/data-structures) tutorial.) In a future lesson, we will learn how to use this value to retrieve the contents of a file.
+The value of the `cid` is a CID object (Content Identifier), a unique address generated from the content of the node. (For a more in-depth look at how CIDs are generated and why they're important, check out our [Decentralized data structures](https://proto.school/#/data-structures) tutorial.) In a future lesson, we will learn how to use this value to retrieve the contents of a file.
```
alanshaw:
```suggestion
The value of the `cid` is a CID object (Content Identifier), a unique address generated from the contents of each file. (For a more in-depth look at how CIDs are generated and why they're important, check out our [Decentralized data structures](https://proto.school/#/data-structures) tutorial.) In a future lesson, we will learn how to use this value to retrieve the contents of a file.
```

src/tutorials/0005-regular-files-api/04.md
```diff
@@ -3,28 +3,35 @@
...
+const stringContents = bufferedContents.toString() // returns a string
 ```
-Notice how the example above waits for the results of the `cat` method before converting the value to a string. To accomplish that in a single line of code, you'd need to wrap the `await` statement in parentheses like so:
+
+Notice how the example above waits for the concatenation of the results of the `cat` method (using the [`it-concat`](https://www.npmjs.com/package/it-concat) package) before converting the value to a string. To accomplish that in a single line of code, you'd need to wrap the `await` statement in parentheses like so:
 
 ```javascript
-let stringContents = (await ipfs.cat(cid)).toString() // returns a string
+const stringContents = (await concat(ipfs.cat(cid))).toString() // returns a string
 ```
```
alanshaw: I don't think these single line versions add any value to the turorial. Saving a line of code should not be a concern for a newcomer (or arguably any developer).

src/tutorials/0005-regular-files-api/04.md
```diff
@@ -3,28 +3,35 @@
...
 
-When you're ready to try this in the real world, you should note that the `cat` method can result in heavy memory usage, depending on the contents of the file being read. If you find this to be the case, you might want to explore the [catReadableStream](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/FILES.md#catreadablestream) or [catPullStream](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/FILES.md#catpullstream) methods instead.
+When you're ready to try this in the real world, you should note that the above example can result in heavy memory usage, depending on the contents of the file being read. If you find this to be the case, you might want to skip the usage of the `it-concat` package and process each chunk of data iteratively.
+This is the main reason why ipfs returns `Async Iterables` so that we can decide if we want to be aware of memory performance issues or not.
+But in ProtoSchool tutorials, we don't really need to worry about this since our code challenges will use small files, so we can concatenate everything without worrying about performance.
```
alanshaw:
```suggestion
In ProtoSchool tutorials, we don't really need to worry about this since our code challenges use small files, so we can concatenate everything without worrying about performance.
```

src/tutorials/0005-regular-files-api/06.js
```diff
@@ -66,19 +76,21 @@ const validate = async (result, ipfs) => {
...
 
 const code = `/* global ipfs */
+const all = require('it-all')
+
 const run = async (files) => {
   const fileObjectsArray = files.map((file) => { return { path: file.name, content: file }})
-  const addedFiles = await ipfs.add(fileObjectsArray, { wrapWithDirectory: true })
-  const directoryCID = addedFiles[addedFiles.length - 1].hash
+  const addedFiles = await all(ipfs.add(fileObjectsArray, { wrapWithDirectory: true }))
+  const directoryCID = addedFiles[addedFiles.length - 1].cid
```
alanshaw: If we keep the `it-` modules then it might be more appropriate to use `it-last` here.

alanshaw:
> > I'm hesitant to use `it-all`
> 
>     1. Would be amazing to use `toArray()` but don't you think it should be at least stage 3 to make content in ProtoSchool using it? Unless it's a proposal that is popular in the community and has requested it a lot.

No, I agree we should not use `toArray()` yet (and you can't anyway without adding a polyfill) and wasn't suggesting we do 😅 I just wanted to highlight that it'll be possible at some point soon and you'd probably need to do a whole bunch of work to update this again to use `toArray()` instead of `it-all` and remove all your explanations for what it does and why require works in the browser. 

If you stick with `for` loops then you'd not need to do any of that work. I don't want to make things more difficult for students but maybe if the exercises were re-structured in a way that didn't encourage buffering then it wouldn't be any more difficult, and the exercises would not be focused on collecting stuff into arrays.

terichadbourne:
@zebateira Just a heads up that I've merged `code` into your branch, which means the file upload testing is now included.

zebateira:
> I agree we should not use `toArray()` yet (and you can't anyway without adding a polyfill) and wasn't suggesting we do 😅 I just wanted to highlight that it'll be possible at some point soon and when that happens you'll probably want/need to do a whole bunch of work to update this again to use `toArray()` instead of `it-all` and remove all your explanations for what it does and why require works in the browser.
> 
> If you stick with `for` loops then you'd not need to do any of that work. I don't want to make things more difficult for students but maybe if the tasks in the exercises were re-structured then it wouldn't be any more difficult, and the exercises would not be focused on collecting stuff into arrays.

Ah yes, that makes sense! I agree with that. I'll put up some of the most complicated code solutions for us to see the implications of this @terichadbourne.

src/components/Lesson.vue
```diff
@@ -413,14 +414,14 @@ export default {
       // create a sample file for the user to read from, acessible at this CID:
       // QmWCscor6qWPdx53zEQmZvQvuWQYxx1ARRCXwYVE4s9wzJ
       /* eslint-disable no-new */
-      return ipfs.add(this.ipfsConstructor.Buffer.from('You did it!'))
+      return all(ipfs.add(this.ipfsConstructor.Buffer.from('You did it!')))
```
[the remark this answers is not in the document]
zebateira: > You might want to make use of it-last to extract the last item.

Ah so the order is guaranteed then. I couldn't that [documented in the docs](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/FILES.md#add)

src/tutorials/0004-mutable-file-system/05.md
zebateira: The second option looks very interesting! 

@terichadbourne

zebateira: So after discussing this with @terichadbourne, there is a point to be made regarding the output that is shown to the user when they submit the correct solution code.

@alanshaw your proposals would create a difference behavior to the output: ATM _all_ output show the result returned from the solution code.
If we implement either (1) or (2) we would have to have change this behavior.
This inconsistency might cause some confusion to the user. We think it's very important to try and keep code solutions as close to real case scenarios as possible, and so, having the output remain consistent (return statements is what is visualized in the output) is crucial.

I really liked (2) though, but it's better to keep the UX consistent in this case.

src/tutorials/0004-mutable-file-system/05.md
zebateira: We will change the variable to be global just like `ipfs` and leave a note about this to the user.

alanshaw: ooh, that's a good idea!

zebateira:
@alanshaw after discussing this with @terichadbourne we think that it might be better to use the `it-*` packages for the following reasons:

- as far as code changes go, from `all(ipfs.add...` to `ipfs.add(file).toArray()`, it's a one line change, and it will be easier (less LOCs changes to update) than to change the `for` loops to `toArray()`. 
- regarding the contextualization of the user to the `it-*` packages, you are right that it adds some complexity to it, but not that much and later on when we have `toArray()` to use we can remove those small notes about the packages.

So in summary, if we use the `it-*` packages, we are reducing code change in the long run, in order to avoid the for loops which adds some complexity as well to the very complex code solutions of later lessons.

Thank you for the input 🙏 
Hopefully we will be able to change the content to use `toArray()` and everything will be even simpler 🙏

src/components/Lesson.vue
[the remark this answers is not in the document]
alanshaw: Not sure about ordering guarantees but by virtue of constructing a DAG the last item _will_ be the root node. So for a number of files wrapped in a directory the last item will be the directory. For a single item (like we have here) the last (and first) item is what you want to extract.

src/tutorials/0004-mutable-file-system/05.md
alanshaw: Understood - perhaps consider re-wording "If we wanted to view all available details for the contents of a `/catPics`" to not say "view"...? You'll get to "view" them because that's what protoschool does for you, but _you_ are not writing code to "view", you're just listing the contents and returning them.

zebateira: @alanshaw yeah good point.
Maybe we could re-word it to "inspect" the directory contents?

src/tutorials/0005-regular-files-api/06.js
zebateira: I think it is ok to keep it as is since we only have one extra LOC and we don't really envision any performance issues in the tutorials.

What do you think @terichadbourne?

src/tutorials/0004-mutable-file-system/10.md
```diff
@@ -8,23 +8,20 @@ MFS has a [`files.read`](https://github.com/ipfs/interface-js-ipfs-core/blob/mas
...
 
 ```js
-await ipfs.files.read(path, [options])
+ipfs.files.read(path, [options])
 ```
 
 The `path` provided is the path of the file to read, and it must point to a file rather than a directory.
 
-The `files.read` method returns a Buffer, which can be converted to a string using the method `toString('utf8')`. For example:
+The `files.read` method returns an Async Iterable, that iterates over the file's chunks of data, i.e. Buffers, which can be converted to a string using the method `toString('utf8')`. For example:
```
terichadbourne:
```suggestion
The `files.read` method returns an Async Iterable that iterates over the file's chunks of data, i.e. Buffers, which can be converted to a string using the method `toString('utf8')`. For example:
```

src/tutorials/0004-mutable-file-system/10.md
```diff
@@ -8,23 +8,20 @@ MFS has a [`files.read`](https://github.com/ipfs/interface-js-ipfs-core/blob/mas
...
 
 ```js
-let bufferedContents = await ipfs.files.read('/directory/some-file.txt')  // a buffer
-let contents = bufferedContents.toString('utf8') // a string
-```
-
-or
+// the concat variable is globally available (just like ipfs)
```
terichadbourne: We need to introduce `concat` & `it-concat` before using it.

src/tutorials/0004-mutable-file-system/10.md
```diff
@@ -8,23 +8,20 @@ MFS has a [`files.read`](https://github.com/ipfs/interface-js-ipfs-core/blob/mas
...
 
-```js
-let contents = (await ipfs.files.read('/directory/some-file.txt')).toString('utf8') // a string
-// notice the parentheses around the entire await statement
+let bufferedContents = await concat(ipfs.files.read('/directory/some-file.txt'))  // a buffer
+let contents = bufferedContents.toString('utf8') // a string
 ```
 
-When you're ready to try this in the real world, you should note that `files.read` method can result in heavy memory usage based on the contents of the file being read. You might want to explore the [`files.readReadableStream`](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/FILES.md#filesreadreadablestream) or [`files.readPullStream`](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/FILES.md#filesreadpullstream) methods instead.
+When you're ready to try this in the real world, you should note that the above example can result in heavy memory usage, depending on the contents of the file being read. If you find this to be the case, you might want to skip the usage of the `it-concat` package and process each chunk of data iteratively.
```
terichadbourne:
```suggestion
When you're ready to try this in the real world, you should note that the above example can result in heavy memory usage, depending on the contents of the file being read. If you're working with large files and find this to be the case, you might want to skip using the `it-concat` package and instead process each chunk of data iteratively. The main reason IPFS now returns `Async Iterables` is to provide a built-in option for dealing with potential performance issues.
```

src/tutorials/0004-mutable-file-system/10.md
```diff
@@ -8,23 +8,20 @@ MFS has a [`files.read`](https://github.com/ipfs/interface-js-ipfs-core/blob/mas
...
+This is the main reason why ipfs returns `Async Iterables` so that we can decide if we want to be aware of memory performance issues or not.
```
terichadbourne:
```suggestion
```

src/tutorials/0004-mutable-file-system/10.md
```diff
@@ -8,23 +8,20 @@ MFS has a [`files.read`](https://github.com/ipfs/interface-js-ipfs-core/blob/mas
...
+In ProtoSchool tutorials, we don't really need to worry about this since our code challenges use small files, so we can concatenate everything without worrying about performance.
```
terichadbourne:
```suggestion
In ProtoSchool tutorials, our code challenges use small files, so we can concatenate everything without worrying about performance.
```

src/tutorials/0004-mutable-file-system/11.js
```diff
@@ -42,51 +50,51 @@ const validate = async (result, ipfs) => {
     return {
       fail: 'Your root directory isn\'t empty. Are you sure you ran the `files.rm` method on your `/some` directory with the option `{ recursive: true }`?',
       logDesc: 'Here are the current contents of your root directory:',
-      log: JSON.stringify(rootDirectoryContents, null, 2)
+      log: rootDirectoryContents.map(utils.format.ipfsObject)
```
terichadbourne: Can you please add this formatting option to the instructions for building tutorials?

src/tutorials/0005-regular-files-api/03.md
```diff
@@ -19,33 +19,54 @@ When you add a file to IPFS, you're putting it only in your own node, but making
...
 
 Note that the `add` method can accept either a single `File` or an array, just in case you have more than one adorable animal photo to add to the node:
 
 ```javascript
-await ipfs.add([catPic, dogPic, giraffePic])
+ipfs.add([catPic, dogPic, giraffePic])
 ```
 
-Because the `add` method returns a `Promise`, you'll need to place an `await` before the method call to suspend the execution until the return value of the promise is ready to be used.
+Because the `add` method returns an [`Async Iterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of), you can only iterate over the values, one by one. In the case you need to return all the values, you can save each one into an array and return it in the end.
```
terichadbourne:
```suggestion
Because the `add` method returns an [`Async Iterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of), you can only iterate over the values, one by one. If you need to return all the values, you can save each one into an array and then return the array.
```

src/tutorials/0005-regular-files-api/03.md
```diff
@@ -19,33 +19,54 @@ When you add a file to IPFS, you're putting it only in your own node, but making
...
 
-The result of this `Promise` is an array of objects, one for each file added to IPFS, in the following format:
+For this, we can use a `for await...of` loop:
```
terichadbourne:
```suggestion
To iterate through the Async Iterable, we can use a `for await...of` loop:
```

src/tutorials/0005-regular-files-api/04-exercise.md
```diff
@@ -7,3 +7,5 @@ QmWCscor6qWPdx53zEQmZvQvuWQYxx1ARRCXwYVE4s9wzJ
 Using the `cat` function, retrieve the file's contents and return them as a string.
 
 **Hint:** The `CID` we provide is a string, and as such it be placed inside quotes. Also, don't forget to convert the buffered contents of the text file to a string using `.toString()` before returning your result.
```
terichadbourne:
```suggestion
**Hint:** The `CID` we provide is a string, so it needs to be placed inside quotes. Also, don't forget to convert the buffered contents of the text file to a string using `.toString()` before returning your result.
```

src/tutorials/0005-regular-files-api/04.md
```diff
@@ -3,28 +3,29 @@
...
 
 The `cat` method first searches your own node for the file requested, and if it can't find it there, it will attempt to find it on the broader IPFS network. Because of the way cryptographic hashing works, the content you're searching for is guaranteed to match the file found, regardless of which peer is hosting it.
 
-Data returned by the `cat` method comes in the form of a `Buffer`. A `Buffer` is just a raw collection of bytes and as such doesn't make any assumptions about encodings or the type of data it contains. However, if we know the file being retrieved is a plain text file such as a `.txt`, we can convert its buffered contents to a UTF-8 string (an interpretation of those raw bytes) by calling the JavaScript method `.toString()`.
+The `cat` method returns an Async Iterable, that iterates over the file's chunks of data, i.e. Buffers.
```
terichadbourne:
```suggestion
The `cat` method returns an Async Iterable that iterates over the file's chunks of data, i.e. Buffers.
```

src/tutorials/0005-regular-files-api/04.md
```diff
@@ -3,28 +3,29 @@
...
+const stringContents = bufferedContents.toString() // returns a string
 ```
 
-When you're ready to try this in the real world, you should note that the `cat` method can result in heavy memory usage, depending on the contents of the file being read. If you find this to be the case, you might want to explore the [catReadableStream](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/FILES.md#catreadablestream) or [catPullStream](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/FILES.md#catpullstream) methods instead.
+When you're ready to try this in the real world, you should note that the above example can result in heavy memory usage, depending on the contents of the file being read. If you find this to be the case, you might want to skip the usage of the `it-concat` package and process each chunk of data iteratively.
+This is the main reason why ipfs returns `Async Iterables` so that we can decide if we want to be aware of memory performance issues or not.
+In ProtoSchool tutorials, we don't really need to worry about this since our code challenges use small files, so we can concatenate everything without worrying about performance.
```
terichadbourne:
```suggestion
In ProtoSchool tutorials, our code challenges use small files, so we can concatenate everything without worrying about performance.
```

src/tutorials/0005-regular-files-api/06.md
```diff
@@ -23,7 +23,7 @@ The `ls` method in the Regular Files API returns an array of objects, one for ea
 
 ```javascript
 {
-  "hash": string,
+  "cid": Object,
```
terichadbourne: Take a look at capitalization througout

src/tutorials/0004-mutable-file-system/05.md
```diff
@@ -23,20 +23,38 @@ such as `/catPics`,
 `files.ls` produces an array of objects, one for each file or directory
 contained in the directory you're inspecting, with the following properties:
 
-- `name` (default): the file's name
-- `type` (optional with `{ long: true }`): the object's type (`0` - file or `1` - directory)
-- `size` (optional with `{ long: true }`): the size of the file in bytes
-- `hash` (optional with `{ long: true }`): the cryptographic hash, or Content Identifier (CID), that represents your file in IPFS
+- `name`: the file's name
+- `type`: the object's type (`0` - file or `1` - directory)
```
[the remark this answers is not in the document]
zebateira: So I just checked with @achingbrain and the docs needs an update on this. Will link the PR here.

src/tutorials/0004-mutable-file-system/05.md
```diff
@@ -23,20 +23,38 @@ such as `/catPics`,
...
 directory, we could do it like this:
 
 ```js
-await ipfs.files.ls('/catPics', { long: true })
+ipfs.files.ls('/catPics')
+```
+
+Because the `files.ls` method returns an [`Async Iterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of), you can only iterate over the values, one by one. In the case you need to return all the values, you can save each one into an array and return it in the end.
+
+For this, we can use a `for await...of` loop:
```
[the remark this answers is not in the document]
zebateira: I'd go for `To iterate over all the values, we can...` because we actually interate over the values that the iterable will provide.

src/tutorials/0005-regular-files-api/03.md
zebateira: Same as the previous one

zebateira:
@terichadbourne I accepted most of the changes you proposed and will work on the ones you requested.
Looking good 👍

src/tutorials/0004-mutable-file-system/05.md
```diff
@@ -23,20 +23,36 @@ such as `/catPics`,
...
 directory, we could do it like this:
 
 ```js
-await ipfs.files.ls('/catPics', { long: true })
+ipfs.files.ls('/catPics')
+```
+
+Because the `files.ls` method returns an [`Async Iterable`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of), you can only iterate over the values, one by one. If you need to return all the values, you can save each one into an array and then return the array.
+
+For this, we can use a `for await...of` loop:
```
zebateira:
```suggestion
To iterate over all the values, we can use a `for await...of` loop:
```

src/tutorials/0004-mutable-file-system/10.md
zebateira: updated here: https://github.com/ProtoSchool/protoschool.github.io/pull/381/commits/771fa666432f40e0f4363d0d695bd6246db5e610

src/tutorials/0004-mutable-file-system/11.js
zebateira: added

src/tutorials/0004-mutable-file-system/05.md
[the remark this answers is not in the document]
zebateira: PR here: https://github.com/ipfs/js-ipfs/pull/2939

DEVELOPING_TUTORIALS.md
```diff
@@ -135,8 +135,8 @@ Follow the steps below to create each lesson.
 
 Depending on which lesson format you've chosen, you'll need to create 2-4 files within your project directory. Check the table below to see which files you need, then read on for instructions on how to create them.
 
-| File | Sample Filename | Standard Lesson with Coding Exercise | Lesson with Coding Exercise and File Upload | Multiple-Choice Lesson | Text-Only Lesson
-| :--- | :---: | :---: | :---: | :---: | :---: |
+| File | Sample Filename | Standard Lesson with Coding Exercise | Lesson with Coding Exercise and File Upload | Multiple-Choice Lesson | Text-Only Lesson |
+| ---- | ---- | ---- | ---- | ---- | ---- |
```
terichadbourne:
```suggestion
| :---  | ---- | ---- | ---- | ---- | ---- |
```

DEVELOPING_TUTORIALS.md
```diff
@@ -388,6 +387,17 @@ You may (optionally) use [markdown formatting](https://guides.github.com/feature
 
 If this is the last lesson in your tutorial, please create a success message that notes that the user has completed the whole tutorial. For example, `Great job! You've completed this series of lessons!`)
 
+###### Utils module
+
+There are some utils that you can use accross lesson code validations. For that you can use the `utils` module:
+
+- `utils.format`: format specific objects like `ipfs` objects
```
terichadbourne:
```suggestion
- `utils.format`: format specific objects, such as `ipfs` objects
```
When and why would I do this?

DEVELOPING_TUTORIALS.md
```diff
@@ -388,6 +387,17 @@ You may (optionally) use [markdown formatting](https://guides.github.com/feature
...
+- `utils.ipfs`: ipfs specific helpers. At the moment includes the error codes that `js-ipfs` might return.
```
terichadbourne:
```suggestion
- `utils.ipfs`: IPFS-specific helpers. Includes the error codes that `js-ipfs` might return.
```

DEVELOPING_TUTORIALS.md
```diff
@@ -602,6 +612,73 @@ The properties exported from this file depend on the lesson type, as follows:
...
+  }
+  ```
+
+- `utils.ipfs`
+
+  - `errorCodes`: set of error codes from `js-ipfs`
+    - `ERR_MORE_THAN_ONE_ROOT`
+
+  ```js
+  if (utils.ipfs.errorCodes.ERR_MORE_THAN_ONE_ROOT) {
```
terichadbourne: problem with code sample

DEVELOPING_TUTORIALS.md
```diff
@@ -602,6 +612,73 @@ The properties exported from this file depend on the lesson type, as follows:
   - `question`: String - mandatory - _the question to be answered by the user_ ([detailed docs](#create-multiple-choice-quizzes-in-your-javascript-file-skip-for-coding-exercises-and-text-only-lessons))
   - `choices`: Array - mandatory - _the potential answers between which the user must choose_ ([detailed docs](#create-multiple-choice-quizzes-in-your-javascript-file-skip-for-coding-exercises-and-text-only-lessons))
 
+### `utils` module
+
+This module is a set of utils designed to be re-used accross the validation code of different tutorials.
+
+- `utils.format`
+
+  - `ipfsObject(object)`: formats an ipfs object to be ready to print in the output ui, for example, formatting a `cid` object into a string.
```
terichadbourne: need examples of what the user sees when you don't use this and when you do

terichadbourne:
```suggestion
  - `ipfsObject(object)`: formats an IPFS object to be ready to print in the output UI, for example, formatting a `cid` object into a string.
```

DEVELOPING_TUTORIALS.md
terichadbourne: When do I use this? How does it relate to the override feature?

DEVELOPING_TUTORIALS.md
```diff
@@ -602,6 +612,73 @@ The properties exported from this file depend on the lesson type, as follows:
...
+    return {
+      fail: 'Some error message specific to this edge case.'
+    }
+  }
+  ```
+
+  
+
+- `utils.validationMessages`
+
```
terichadbourne: Add context that these produce messages shown to the user - author still has to built the code that does the testing

DEVELOPING_TUTORIALS.md
```diff
@@ -602,6 +612,73 @@ The properties exported from this file depend on the lesson type, as follows:
...
+  ```
+
+  
+
+- `utils.validators`
+
+  - `isAsyncIterable(result)`: tests whether the result is an AsyncIterable or not
+
+  ```js
+  if (utils.validators.isAsyncIterable(result)) {
```
terichadbourne: condition wrong

DEVELOPING_TUTORIALS.md
terichadbourne: show an example of messages shown to user

src/tutorials/utils/format.js
```diff
@@ -1,3 +1,8 @@
+/*
+  Format an ipfs object. Changes applied ATM:
```
terichadbourne:
```suggestion
  Format an IPFS object. Changes applied ATM:
```

src/tutorials/utils/index.js
```diff
@@ -1,3 +1,12 @@
+/*
+  Set of utils to help build validation code
+
+  - format: format data for showing in the output ui
```
terichadbourne:
```suggestion
  - format: format data to display in the output UI
```

src/tutorials/utils/index.js
```diff
@@ -1,3 +1,12 @@
...
+  - ipfs: js-ipfs related helpers
```
terichadbourne:
```suggestion
  - ipfs: helpers related to js-ipfs
```

src/tutorials/utils/ipfs.js
```diff
@@ -1,3 +1,6 @@
+/*
+  Error codes strings used in IPFS
```
terichadbourne:
```suggestion
  Error code strings used in IPFS
```

src/tutorials/utils/validation-messages.js
```diff
@@ -1,3 +1,6 @@
+/*
+  Error messages to show to the use on the output ui
```
terichadbourne:
```suggestion
  Error messages to show to the use on the output UI
```

DEVELOPING_TUTORIALS.md
```diff
@@ -602,6 +612,73 @@ The properties exported from this file depend on the lesson type, as follows:
   - `question`: String - mandatory - _the question to be answered by the user_ ([detailed docs](#create-multiple-choice-quizzes-in-your-javascript-file-skip-for-coding-exercises-and-text-only-lessons))
   - `choices`: Array - mandatory - _the potential answers between which the user must choose_ ([detailed docs](#create-multiple-choice-quizzes-in-your-javascript-file-skip-for-coding-exercises-and-text-only-lessons))
 
+### `utils` module
+
+This module is a set of utils designed to be re-used accross the validation code of different tutorials.
+
+- `utils.format`
```
terichadbourne: Switch these to headers instead of bullets so you can link to them from relevant parts of instructions. Reference from other spots in the instructions where the specific util could be helpful.

DEVELOPING_TUTORIALS.md
```diff
@@ -396,7 +396,7 @@ There are some utils that you can use accross lesson code validations. For that
 - `utils.validators`: common validators to be used throughout validation code
 - `utils.validationMessages`: common messages to show the user
 
-For a more detailed docs, see [here](#detailed-docs).
+For a more detailed docs, see [here](#utils-module-1).
```
terichadbourne:
```suggestion
For more detailed docs, see [here](#utils-module-1).
```

src/tutorials/0004-mutable-file-system/10.md
```diff
@@ -13,7 +13,7 @@ ipfs.files.read(path, [options])
 
 The `path` provided is the path of the file to read, and it must point to a file rather than a directory.
 
-The `files.read` method returns an Async Iterable that iterates over the file's chunks of data, i.e. Buffers, which can be converted to a string using the method `toString('utf8')`. For example:
+The `files.read` method returns an Async Iterable that iterates over the file's chunks of data, i.e. Buffers. In our case, we ultimately need to convert Buffers into strings using the method `toString()`. However, the chunks of data within a single file need to be reassembled (concatenated) before the conversion. The [`it-concat`](https://www.npmjs.com/package/it-concat) package can iterate over all of the chunks and put them back together for us:
 
 ```js
 import concat from 'it-concat'
```
terichadbourne:
```suggestion
// the concat function comes from the it-concat package
// and is made globally available (just like ipfs) in our code challenges
```

src/tutorials/0005-regular-files-api/03-exercise.md
```diff
@@ -1,3 +1,5 @@
 Add the files in your browser (available in the `files` array below) to your IPFS node using the `add` method.
 
-**Hint:** You can pass either a single `File` object or an array of `File` objects to the `add` method.
+**Hints:**
+    - You can pass either a single `File` object or an array of `File` objects to the `add` method.
```
terichadbourne:
```suggestion
- You can pass either a single `File` object or an array of `File` objects to the `add` method.
```
It's creating a code block somehow.

src/tutorials/0005-regular-files-api/03-exercise.md
```diff
@@ -1,3 +1,5 @@
...
+    - You need to concatenate all the results into an array because the `ipfs.add` method returns an `Async Iterable`. You can use either the `for await...of` loop or the function `all`.
```
terichadbourne:
```suggestion
 - You need to concatenate all the results into an array because the `ipfs.add` method returns an `Async Iterable`. You can use either the `for await...of` loop or the function `all`.
```

src/tutorials/0005-regular-files-api/03.js
```diff
@@ -38,15 +49,16 @@ const validate = async (result, ipfs) => {
   if (JSON.stringify(expectedResult) === JSON.stringify(result)) {
     return {
       success: 'Success! You did it!',
-      logDesc: 'Your `add` command returned the array of objects below. Notice in particular the `hash` ' + valueText + ", since we'll need " + thatText + ' to access ' + fileText + ' again later. The `path` matches the `hash` for ' + fileText + ", but we'll see in future lessons that that's not always true.",
-      log: result
+      logDesc: 'Your `add` command returned the array of objects below. Notice in particular the `cid` ' + valueText + ", since we'll need " + thatText + ' to access ' + fileText + ' again later. The `path` matches the `cid` for ' + fileText + ", but we'll see in future lessons that that's not always true.",
```
terichadbourne: Probably really matching a field inside the shrunken CID object

src/tutorials/0005-regular-files-api/03.js
```diff
@@ -38,15 +49,16 @@ const validate = async (result, ipfs) => {
...
+      log: result.map(utils.format.ipfsObject)
```
terichadbourne: improve documentation for this

src/tutorials/0005-regular-files-api/04.md
```diff
@@ -3,28 +3,29 @@
...
 
-Data returned by the `cat` method comes in the form of a `Buffer`. A `Buffer` is just a raw collection of bytes and as such doesn't make any assumptions about encodings or the type of data it contains. However, if we know the file being retrieved is a plain text file such as a `.txt`, we can convert its buffered contents to a UTF-8 string (an interpretation of those raw bytes) by calling the JavaScript method `.toString()`.
+The `cat` method returns an Async Iterable that iterates over the file's chunks of data, i.e. Buffers.
+
+A `Buffer` is just a raw collection of bytes and as such doesn't make any assumptions about encodings or the type of data it contains. However, if we know the file being retrieved is a plain text file such as a `.txt`, we can convert its buffered contents to a UTF-8 string (an interpretation of those raw bytes) by calling the JavaScript method `.toString()`.
 
 So if you had the CID for a text file in an IPFS node, you could retrieve the file's contents as a readable string like so:
```
terichadbourne: Looks like we're missing an intro of it-concat here

src/tutorials/0005-regular-files-api/04-exercise.md
```diff
@@ -6,4 +6,6 @@ QmWCscor6qWPdx53zEQmZvQvuWQYxx1ARRCXwYVE4s9wzJ
 
 Using the `cat` function, retrieve the file's contents and return them as a string.
 
-**Hint:** The `CID` we provide is a string, and as such it be placed inside quotes. Also, don't forget to convert the buffered contents of the text file to a string using `.toString()` before returning your result.
+**Hint:** The `CID` we provide is a string, so it needs to be placed inside quotes. Also, don't forget to convert the buffered contents of the text file to a string using `.toString()` before returning your result.
```
terichadbourne: Join the hints!

src/tutorials/0005-regular-files-api/06.md
```diff
@@ -8,13 +8,13 @@ Now that we've wrapped some files in a directory in our IPFS node, let's learn h
 You can call the `ls` method like so:
```
terichadbourne: There's nothing here to introduce the fact that Async Iterables are involved but I got an error message stating I was wrong because of them. :)

src/tutorials/0005-regular-files-api/07-exercise.md
```diff
@@ -14,6 +14,7 @@ Based on our file structure, the path could be expressed as:
 
 There are two valid ways to solve this challenge. Take your pick!
 
-**Hint:** In the code below we've already taken care of converting the `bufferedContents` to a string before returning the result, as you did yourself previously using `.toString()`.
-
-Don't forget to prepend `/ipfs/` to the IPFS Path string
+**Hints:**
+- In the code below we've already taken care of converting the `bufferedContents` to a string before returning the result, as you did yourself previously using `.toString()`.
+- Don't forget to prepend `/ipfs/` to the IPFS Path string
+- You need to concatenate all the data into a single buffer because the `ipfs.cat` method returns an `Async Iterable`. You can use the package `it-concat`.
```
terichadbourne: You never mentioned this in the lesson itself

src/tutorials/0005-regular-files-api/08.md
```diff
@@ -3,17 +3,17 @@
     type: "code"
 ---
```
terichadbourne: You never mentioned async iterables and the samples don't reflect them

src/tutorials/0005-regular-files-api/07-exercise.md
zebateira: This was already addressed in Lesson 4. Should we add more than just the hint then?

src/tutorials/0005-regular-files-api/03.js
zebateira: Even if it's that field, it still represents the CID, so I don't think we need to distinguish here in this case.

zebateira:
@terichadbourne I think I went over all your suggestions now! Let me know if there's anything else missing.

src/tutorials/0005-regular-files-api/07-exercise.md
terichadbourne: No, you're right, sorry. Just the hint is fine.

DEVELOPING_TUTORIALS.md
```diff
@@ -603,6 +614,133 @@ The properties exported from this file depend on the lesson type, as follows:
...
+
+```js
+return {
+  success: 'Success! You did it!',
+  logDesc: "Here is the result.",
+  log: utils.format.ipfsObject(result)
+}
+```
+
+Original `result` object would show like this:
```
terichadbourne:
```suggestion
Without special formatting, the `result` object would be displayed like this:
```

DEVELOPING_TUTORIALS.md
```diff
@@ -603,6 +614,133 @@ The properties exported from this file depend on the lesson type, as follows:
...
+        37
+      ]
+    }
+  },
+  "size": 40391,
+  "mode": 420
+}
+```
+
+While the formatted object would show like this:
```
terichadbourne:
```suggestion
By contrast, the formatted object is displayed like this:
```

DEVELOPING_TUTORIALS.md
```diff
@@ -603,6 +614,133 @@ The properties exported from this file depend on the lesson type, as follows:
...
+  "size": 778308,
+  "mode": 420
+}
+```
+
+#### `utils.ipfs`
+
+- `errorCodes`: set of error codes from `js-ipfs`
+  - `ERR_MORE_THAN_ONE_ROOT`
+
```
terichadbourne:
```suggestion
You can reference these error codes in your validation code when you'd like to override them with messages more specific to the context of your lesson. For example:

```

DEVELOPING_TUTORIALS.md
```diff
@@ -603,6 +614,133 @@ The properties exported from this file depend on the lesson type, as follows:
...
+```js
+if (result.code === utils.ipfs.errorCodes.ERR_MORE_THAN_ONE_ROOT) {
+  return {
+    fail: 'Some error message specific to this edge case.'
+  }
+}
+```
+
```
terichadbourne:
```suggestion
IPFS error codes not listed above are not currently included in this feature, but can be added to our codebase as needed. If the error codes you need are not available, you can add them to the file [tutorials/utils/ipfs.js](./src/tutorials/utils/ipfs.js) and to this documentation.
```

DEVELOPING_TUTORIALS.md
```diff
@@ -603,6 +614,133 @@ The properties exported from this file depend on the lesson type, as follows:
...
+
+
+#### `utils.validationMessages`
+
+These validation messages can be used to avoid writing the same message for the users across tutorials and lessons.
```
terichadbourne:
```suggestion
These validation messages can be used to avoid re-writing the same feedback messages across tutorials and lessons. 
```

DEVELOPING_TUTORIALS.md
```diff
@@ -603,6 +614,133 @@ The properties exported from this file depend on the lesson type, as follows:
...
+- `SUCCESS`: when the user has successfully completed the code challenge
+  - Example: `Success! You did it!`
+- `NO_RESULT`: to show when the user forgets to return a result
+  - Example: `Oops! You forgot to return a result :(`
+- `VALUE_IS_ASYNC_ITERABLE_ALL`: when the user returned an AsyncIterable and forgot to use the `all` function on the result
+  - Example: `The returned value is an Async Iterable. Did you forget to put all the results together using either for await...of or with all?`
+- `VALUE_IS_ASYNC_ITERABLE_CONCAT`: when the user returned an AsyncIterable and forgot to use the `concat` function on the result
+  - Example: `The returned value is an Async Iterable. Did you forget to concatenate all the data using concat?`
+
+
```
terichadbourne: While these codes provide an easy shorthand, you'll still need to provide the logic that determines when a particular validation message should be displayed for each lesson. For example:

DEVELOPING_TUTORIALS.md
```diff
@@ -603,6 +614,133 @@ The properties exported from this file depend on the lesson type, as follows:
...
+```
+
+
+
+#### `utils.validators`
+
+Validators that can be re-used across tutorials and lessons.
+
+- `isAsyncIterable(result)`: tests whether the result is an AsyncIterable or not
+
```
terichadbourne:
```suggestion

While you can use these shorthands to test for certain conditions, you'll still need to specify what validation message should be displayed as a result. In the example below, a validator is used in combination with a validation message. 
```

DEVELOPING_TUTORIALS.md
```diff
@@ -388,6 +387,17 @@ You may (optionally) use [markdown formatting](https://guides.github.com/feature
 
 If this is the last lesson in your tutorial, please create a success message that notes that the user has completed the whole tutorial. For example, `Great job! You've completed this series of lessons!`)
```
terichadbourne: [](url)
```suggestion
If this is the last lesson in your tutorial, please create a success message that notes that the user has completed the whole tutorial. For example, `Great job! You've completed this series of lessons!`)

You can optionally use the [utils.validationMessages](#utilsvalidationmessages) utility to create commonly used validation messages. 
```

terichadbourne: I'd love to add in-line mentions for every common use case of each of the util functions so people find them in context.

src/tutorials/0005-regular-files-api/04.md
```diff
@@ -17,6 +17,8 @@ The `cat` method returns an Async Iterable that iterates over the file's chunks
 
 A `Buffer` is just a raw collection of bytes and as such doesn't make any assumptions about encodings or the type of data it contains. However, if we know the file being retrieved is a plain text file such as a `.txt`, we can convert its buffered contents to a UTF-8 string (an interpretation of those raw bytes) by calling the JavaScript method `.toString()`.
 
+But since we have multiple buffers of the same file, we need to reassemble them (concatenate) into a single buffer before converting it into a string. The [`it-concat`](https://www.npmjs.com/package/it-concat) package can do just that: iterate over all of the chunks of the file and put them back together for us.
```
terichadbourne:
```suggestion
But since we have multiple buffers making up a single file, we need to reassemble them (concatenate) into a single buffer before converting it into a string. The [`it-concat`](https://www.npmjs.com/package/it-concat) package can do just that: iterate over all of the chunks of the file and put them back together for us.
```

achingbrain: FWIW `it-concat` turns the buffers into a [`BufferList`](https://www.npmjs.com/package/bl), which you have to call `.slice` on to get an actual `Buffer`.

It's a good approach if you are going to doing lots of operations on the contents afterwards or appending new data, etc but if you want to avoid the conceptual overhead of Buffers vs BufferLists, you can just use [`it-to-buffer`](https://www.npmjs.com/package/it-to-buffer) instead.

DEVELOPING_TUTORIALS.md
zebateira:
```suggestion
While these codes provide an easy shorthand, you'll still need to provide the logic that determines when a particular validation message should be displayed for each lesson. For example:
```

src/tutorials/0005-regular-files-api/04.md
zebateira: hey @achingbrain! Thank you for the input.

That's a good point, we'll switch to `it-to-buffer` then 👍 
Also, I just noticed that `it-concat` depends on [`bl`](https://www.npmjs.com/package/bl), so maybe it is intended for nodejs and not browers? @alanshaw 
If that's the case, then it's one more reason to switch to `it-to-buffer`.

src/tutorials/0004-mutable-file-system/02.js
```diff
@@ -8,7 +8,12 @@ const validate = async (result, ipfs) => {
   } else if (!!result && result.cid.toString() === 'QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn') {
     return {
       success: utils.validationMessages.SUCCESS,
-      logDesc: "Here's the status of your root directory ( `/` ). Notice that it has a CID even though it doesn't have contents yet. Every empty IPFS node has this exact same CID, because their non-existent contents are identical!",
+      logDesc: [
+        "Here's the status of your root directory (`/`).",
+        `<pre class="code-highlight"><code class="hljs json">${JSON.stringify(result, null, 2)}</code></pre>`,
+        'The output is very long because the CID is represented as an `Object` internally. To simplify, this we can use the `toString()` method on the `cid` property to get the CID in string format: `' + result.cid.toString() + '`. In future lessons we\'ll always show this simplified version to make it easier to read as shown bellow. <br/> <br/>',
+        "Notice that it has a CID even though it doesn't have contents yet. Every empty IPFS node has this exact same CID, because their non-existent contents are identical!"
```
terichadbourne, line 15:
```suggestion
        "Notice that our node has a CID even though it doesn't have contents yet. Every empty IPFS node has this exact same CID, because their non-existent contents are identical!"
```

src/tutorials/0004-mutable-file-system/10.md
```diff
@@ -13,14 +13,14 @@ ipfs.files.read(path, [options])
 
 The `path` provided is the path of the file to read, and it must point to a file rather than a directory.
 
-The `files.read` method returns an Async Iterable that iterates over the file's chunks of data, i.e. Buffers. In our case, we ultimately need to convert Buffers into strings using the method `toString()`. However, the chunks of data within a single file need to be reassembled (concatenated) before the conversion. The [`it-concat`](https://www.npmjs.com/package/it-concat) package can iterate over all of the chunks and put them back together for us. (We've made this package availabe to you in our excercises as `concat`.)
+The `files.read` method returns an Async Iterable that iterates over the file's chunks of data, i.e. Buffers. In our case, we ultimately need to convert Buffers into strings using the method `toString()`. However, the chunks of data within a single file need to be reassembled (concatenated) before the conversion. The [`it-to-buffer`](https://www.npmjs.com/package/it-to-buffer) package can iterate over all of the chunks and put them back together for us. (We've made this package availabe to you in our excercises as `concat`.)
```
terichadbourne, line 16:
```suggestion
The `files.read` method returns an Async Iterable that iterates over the file's chunks of data, i.e. Buffers. In our case, we ultimately need to convert Buffers into strings using the method `toString()`. However, the chunks of data within a single file need to be reassembled (concatenated) before the conversion. The [`it-to-buffer`](https://www.npmjs.com/package/it-to-buffer) package can iterate over all of the chunks and put them back together for us. (We've made this package availabe to you in our exercises as `toBuffer`.)
```

src/tutorials/0004-mutable-file-system/02.js
```diff
@@ -8,7 +8,12 @@ const validate = async (result, ipfs) => {
   } else if (!!result && result.cid.toString() === 'QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn') {
     return {
       success: utils.validationMessages.SUCCESS,
-      logDesc: "Here's the status of your root directory ( `/` ). Notice that it has a CID even though it doesn't have contents yet. Every empty IPFS node has this exact same CID, because their non-existent contents are identical!",
+      logDesc: [
+        "Here's the status of your root directory (`/`).",
+        `<pre class="code-highlight"><code class="hljs json">${JSON.stringify(result, null, 2)}</code></pre>`,
+        'The output is very long because the CID is represented as an `Object` internally. To simplify, this we can use the `toString()` method on the `cid` property to get the CID in string format: `' + result.cid.toString() + '`. In future lessons we\'ll always show this simplified version to make it easier to read as shown bellow. <br/> <br/>',
```
terichadbourne, line 14:
```suggestion
        'The output is very long because the CID is represented as an `Object` internally. To simplify, this we can use the `toString()` method on the `cid` property to get the CID in string format: `' + result.cid.toString() + '`. In future lessons we\'ll always show this simplified version to make it easier to read, as shown below. <br/> <br/>',
```

src/tutorials/0005-regular-files-api/03.js
```diff
@@ -49,7 +49,12 @@ const validate = async (result, ipfs) => {
   if (JSON.stringify(expectedResult) === JSON.stringify(result)) {
     return {
       success: utils.validationMessages.SUCCESS,
-      logDesc: 'Your `add` command returned the array of objects below. Notice in particular the `cid` ' + valueText + ", since we'll need " + thatText + ' to access ' + fileText + ' again later. The `path` matches the `cid` for ' + fileText + ", but we'll see in future lessons that that's not always true.",
+      logDesc: [
+        'The output is very long because the CID is represented as an `Object` internally.',
+        `<pre class="code-highlight"><code class="hljs json">${JSON.stringify(result, null, 2)}</code></pre>`,
+        'To simplify, this we can use the `toString()` method on the `cid` property to get the CID in string format: `QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn`. In future lessons we\'ll always show this simplified version to make it easier to read as shown bellow. <br/> <br/>',
```
terichadbourne, line 55:
```suggestion
        'To simplify, this we can use the `toString()` method on the `cid` property to get the CID in string format: `QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn`. In future lessons we\'ll always show this simplified version to make it easier to read, as shown below. <br/> <br/>',
```

src/tutorials/0005-regular-files-api/03.js
```diff
@@ -49,7 +49,12 @@ const validate = async (result, ipfs) => {
   if (JSON.stringify(expectedResult) === JSON.stringify(result)) {
     return {
       success: utils.validationMessages.SUCCESS,
-      logDesc: 'Your `add` command returned the array of objects below. Notice in particular the `cid` ' + valueText + ", since we'll need " + thatText + ' to access ' + fileText + ' again later. The `path` matches the `cid` for ' + fileText + ", but we'll see in future lessons that that's not always true.",
+      logDesc: [
+        'The output is very long because the CID is represented as an `Object` internally.',
```
terichadbourne, line 53:
```suggestion
        'Your `add` command returned the array of objects below. The output is very long because the CID is represented as an `Object` internally, but if you scroll down we'll offer you a more condensed view.',
```

src/tutorials/0004-mutable-file-system/02.js
```diff
@@ -8,7 +8,12 @@ const validate = async (result, ipfs) => {
   } else if (!!result && result.cid.toString() === 'QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn') {
     return {
       success: utils.validationMessages.SUCCESS,
-      logDesc: "Here's the status of your root directory ( `/` ). Notice that it has a CID even though it doesn't have contents yet. Every empty IPFS node has this exact same CID, because their non-existent contents are identical!",
+      logDesc: [
+        "Here's the status of your root directory (`/`).",
```
terichadbourne, line 12:
```suggestion
        "Here's the status of your root directory (`/`). The output is very long because the CID is represented as an `Object` internally, but if you scroll down we'll offer you a more condensed view.",
```

src/tutorials/0004-mutable-file-system/02.js
terichadbourne:
```suggestion
        'To simplify the output, we can use the `toString()` method on the `cid` property to get the CID in string format: `' + result.cid.toString() + '`. In future lessons we\'ll always show this simplified version to make it easier to read, as shown below. <br/> <br/>',
```

PR #384, opened 2020-02-17
ipfs@0.41 update: option 1

zebateira:
First option for dealing with the new js ipfs api that always returns iterables.
This first approach adds a `for await...of` loop.

Screen Shot 2020-02-17 at 7 04 37 PM

- Still working out how to solve the linting warnings on the monaco editor
- Although it looks a bit convoluted, it might be a better choice if we don't want to introduce a module when teaching

See original WIP PR: #381

### DIFF
--- a/src/tutorials/0005-regular-files-api/03.js
+++ b/src/tutorials/0005-regular-files-api/03.js
@@ -1,7 +1,10 @@
 const validate = async (result, ipfs) => {
   const uploadedFiles = window.uploadedFiles || false
+  const expectedResult = []
 
-  const expectedResult = await ipfs.add(window.uploadedFiles)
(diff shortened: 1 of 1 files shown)


zebateira:
Alternatively, we could also take the `ipfs.add` operation out of the for loop:

Initial code:
```js
/* global ipfs */
  const result = []
  const operation = // Place your code to add a file or files here

  for await (const resultPart of operation) {
    result.push(resultPart)
  }

  return result
}
return run
```

Solution code:
```js
/* global ipfs */
  const result = []
  const operation = ipfs.add(files)

  for await (const resultPart of operation) {
    result.push(resultPart)
  }

  return result
}
return run
```

PR #389, opened 2020-02-20
Add Cypress tests for multiple choice lessons

### LINKED ISSUE #390: Create Cypress tests for multiple-choice lessons
Need to add tests for multiple choice lesson types, which would be great to have in place before launching our first tutorial of that type (#359) 

Work is underway in #389 but we need to first merge #378.

terichadbourne:
will close #390 by way of #359 once that one's merged

@zebateira I'm almost done with this, but I hit a formatting snag when it tries to check that the answers and feedback for each choice match what's expected. It's only breaking in specific cases, which leads me to believe it's about a specific combo of quote types. 

I tried to fix it by importing `marked` and applying it the way it's applied in the Output component, but I hit an error that looked like maybe the module wasn't loading correctly (see commented out code for what I tried). Could you help me debug so we can make this comparison work?

The specific strings that break it are as follows (choice numbers are the index within the choices array): 

 lesson 01 WRONG choice 0 ANSWER 
     CypressError: Timed out retrying: expected '<label>' to contain '`sha1`'
      
lesson 01 WRONG choice 2 ANSWER 
     CypressError: Timed out retrying: expected '<label>' to contain 'It uses a combination of `sha1`, `sha2-256`, `sha3-256`, `sha3-512` and `shake-256`.'

lesson 01 RIGHT choice 1 ANSWER 
     CypressError: Timed out retrying: expected '<label>' to contain '`sha2-256`'

lesson 02 WRONG choice 0 ANSWER
     CypressError: Timed out retrying: expected '<label>' to contain 'The length of the hash tells us which cryptographic algorithm was used. If the length is `256` then the algorithm used was `sha2-256`. If the length is `512` then the algorithm used was `sha2-512`.'
      
 lesson 02 WRONG choice 0 FEEDBACK 
     CypressError: Timed out retrying: expected '<div.output-log.lh-copy.bg-red.white>' to contain 'It\'s true that the length of a `sha2-256` hash is `256`, but this isn\'t enough information to tell us what hashing algorithm was used. Check the information above and try again.'
     
lesson 05 WRONG choice 0 ANSWER
     CypressError: Timed out retrying: expected '<label>' to contain 'We have to guess it, but it\'s easy because most implementations of IPFS encode CIDs with `base32`.'

 lesson 05 WRONG choice 1 ANSWER
     CypressError: Timed out retrying: expected '<label>' to contain 'We can check the initial characters of the hash. If it starts with `Qm`, it\'s `base58btc`. If it starts with `b` it\'s `base32`. In both cases the CID version is `CIDv1`.'
 
lesson 05 RIGHT choice 2 ANSWER 
     CypressError: Timed out retrying: expected '<label>' to contain 'We can check the initial characters of the hash. If it starts with `Qm`, it\'s `base58btc` and is `CIDv0`. If it starts with `b` it\'s `base32` and is `CIDv1`.'


zebateira:
@terichadbourne fixed!
I removed the markdown processing and simply removed the "`" character from the output and it seems to work fine. Let me know if this works for the tests.

cypress/integration/tutorials.spec.js
```diff
@@ -1,79 +1,229 @@
...
+const testLessons = {
+  code: {
+    tutorialId: '0002',
+    lessonNr: '02'
+  },
+  fileUpload: {
+    tutorialId: '0004',
+    lessonNr: '05'
+  },
+  // TODO: Uncomment when we have testing for multiple choice built AND Anatomy of a CID is published
```
zebateira: this can be removed?

cypress/integration/tutorials.spec.js
```diff
@@ -1,79 +1,229 @@
...
+    cy.visit(`/#/tutorials/`)
+    cy.get('[data-cy=tutorial-title]').should('have.length', courses.all.length) // displaying # of tutorials in all array in courses.json
+    cy.get('[data-cy=tutorial-title]').should('have.length', Object.keys(tutorials).length) // displaying # of tutorials in tutorials.json
+    for (let i = 0; i < courses.all.length; i++) {
+      cy.get('[data-cy=tutorial-title]').eq(i).should('contain', tutorials[courses.all[i]].title)
+    }
+  })
+  it(`toggle hides coding tutorials`, function () {
+    const codelessTutorials = courses.all.filter(tutorialId => (getTutorialType(tutorialId) !== 'code') && (getTutorialType(tutorialId) !== 'file-upload'))
+    cy.log('codelessTutorials', codelessTutorials)
```
zebateira: can we remove?

cypress/integration/tutorials.spec.js
```diff
@@ -1,79 +1,229 @@
...
-  standardCodingTutorials.forEach(tutorialId => {
-    describe(`pass ${tutorialId}`, function () {
-      viewSolutionsAndSubmitAll(tutorialId)
+describe(`ADVANCES THROUGH ALL LESSONS IN ALL TUTORIALS`, function () {
+  Object.keys(tutorials).forEach(tutorialId => {
+    describe(`advances through tutorial ${tutorialId} (${tutorials[tutorialId].url})`, function () {
+      advanceThroughLessons(tutorialId)
     })
   })
+  // SWAP OUT ABOVE/BELOW TO MAKE MULT CHOICE TESTING FASTER BY SKIPPING OTHER TUTORIALS
```
zebateira: can we remove this?

zebateira:
@terichadbourne I'm guessing you meant to create the PR to merge into #378 ?

terichadbourne:
Okay, I've merge `code` into this (which included the other testing updates that were recently merged into there) and made a couple of tweaks based on your suggestions. 

I think our options are: 
- Remove the directory for your Anatomy of a CID tutorial and merge this into code - note that this would bring with it some edits other edits you've made in your PR that are not in the lesson files - looks like there are some things there that affect logo display, a couple of typo fixes in other lessons, etc. 
- Attempt to remove from this branch all of the updates that are in the Anatomy of a CID branch that aren't already in code so we're merging purely the testing updates now into code
- Merge this into your tutorial branch - it won't merge until that tutorial is ready, which is okay since I don't anticipate anyone else building multiple choice lessons before then - but we'd have a fairly multi-purposed PR
- Sit on this and merge it on the same day we merge your PR (in the meantime they'll grow further apart from each other with any content updates to the tutorial)

Do any of these strike you as the best option?

terichadbourne:
Update: I'm working on adding a new test to this to check the thing that's currently broken in the Anatomy of a CID branch, where it's getting stuck on `lessonPassed` status even if you switch back to a wrong answer. Currently the test always checks the correct answer last so it wouldn't have been caught.

zebateira:
> Do any of these strike you as the best option?

I'd say the last option: let's just wait and merge it when #329 is merged. We're almost done with it and this one as well, so I think it's fine.

cypress/integration/tutorials.spec.js
```diff
@@ -13,7 +11,7 @@ const testLessons = {
     tutorialId: '0004',
     lessonNr: '05'
   },
-  // TODO: Uncomment when we have testing for multiple choice built and Anatomy of a CID is published
+  // TODO: Uncomment when Anatomy of a CID is published
```
zebateira: Can we use this now?

terichadbourne:
@zebateira I've refactored to test whether the lessonPassed status changes correctly when you flip from a right answer to a wrong one. The test in this branch fails where it should if that feature isn't working.  (It tests in detail that for a specific test case, then only uses the correct choice when it loops through every lesson.)

I still can't get that multiple choice functionality to work correctly in your Anatomy of a CID branch, so I can't merge the solution in here to test it. But looking under the hood, the code here already includes the edits you made in this commit that supposedly made the output reactive.  https://github.com/ProtoSchool/protoschool.github.io/pull/359/commits/559d62c3239a8e56805219d4679f4b037e27801b

My interpretation is that my test is correctly catching an error in our code, but I know you think you've solved that problem locally so I'm not sure where the discrepancy is between what we're seeing. Actually, it's not the Output component that was failing for me, it was the Progress component, so maybe you fixed something different from the issue I was raising?  ¯\_(ツ)_/¯

zebateira:
> so maybe you fixed something different from the issue I was raising?

Yeah we were talking about different issues! Both are fixed now:

- Output not reacting to success or failure, only the first response
- Progress message not updated on failure

terichadbourne:
@zebateira I believe this one is ready for review and is now set into your Anatomy of a CID branch, which means that it will only close #390 after your PR #359 gets merged (which is fine since there won't be multiple choice lessons to test until then. 

Mind giving it a final review?

PR #392, opened 2020-02-20
fix: use ipld format instead of multicodec identifiers

terichadbourne:


terichadbourne:
This is small enough that I think it'll be easier to go ahead and merge it so folks can review the whole tutorial with the suggested edits.

Issue #258, opened 2019-07-10
Feature: Enable separate metadata for different SPA page views for social sharing 


terichadbourne:
Although analytics via Countly (PR #257) will successful capture page views separately for different views/routes in our SPA, we still have the challenge of only being able to have a single set of metadata for things like social card generation. Opening a new issue here as a spinoff of #179 which will be closed shortly. 

Shared by @mikeal: we could use this to get the right card info (open graph metadata) on different pages https://github.com/chrisvfritz/prerender-spa-plugin

Issue #105, opened 2018-12-18
Nicer URLs :)


mikeal:
Currently, we're using hash based routing in Vue.js, which means that the URLs aren't as pretty as they could be :)

We could us Vue.js' [history mode](https://router.vuejs.org/guide/essentials/history-mode.html) but it's a little difficult to make it work with ghpages and/or Cloudflare.

One option would be to use history mode *and* create static pages for each resources, https://cli.vuejs.org/config/#pages. This would greatly improve our SEO but might increase our build times a bit.

Frijol:
Was this resolved (possibly in https://github.com/ProtoSchool/protoschool.github.io/pull/103)? URLs appear to be very sensible from a user perspective at the moment

terichadbourne:
@mikeal Was your "not pretty" assessment just related to the URLs all including "/#/" or did you have something else in mind as well?  Not that this convo will be affected by whether or not we move to VuePress (see #141).

terichadbourne:
@zebateira 

Note that these issues are probably all affected in some way by the choices we make here:
- Want to be able to create separate meta tags for social sharing on separate "pages" (#258)
- We haven't been able to use anchor tags to drop people down to specific content within a page because of the hash routing 
- We need to think about what will work best longer term when we're ready to host on IPFS (#18)

terichadbourne:
Closing as @zebateira has just gotten us switched over to history mode (removing the hashes and setting up redirects) and transitioned to Netlify. This should unblock the SEO / metadata challenges (#258), enable anchor tags, and help with a future transition to IPFS & Fleek (#18). 🎉

PR #393, opened 2020-02-21
Use mode: history on the vue-router by using a SPA hack for Github Pages

### LINKED ISSUE #388: Bug: Loading GitHub default 404
Just spotted this note from @zebateira on an older PR that's already been merged. When we changed the routing I believe we created fewer cases in which we can load our own 404, so we'll need to set up an alternative via GitHub for cases where, for example, the `#` is omitted from the URL. 

> Ah nice catch!
> 
>  _Interestingly hitting a GitHub-generated 404 and not our own - will this case be fixed by the routing PR you have in progress by any chance?_
> 
> hm no, unfortunately, when you access any route besides / you will get github's 404.
> but we can customize it using: https://help.github.com/en/github/working-with-github-pages/creating-a-custom-404-page-for-your-github-pages-site

zebateira:
This PR implements a hack that enables non-hashing routing to github-pages-powered websites. It's working locally in dev and production mode.

Hack can be found here: https://github.com/rafrex/spa-github-pages

SEO Note from the repo:

> A quick SEO note - while it's never good to have a 404 response, it appears based on Search Engine Land's testing that Google's crawler will treat the JavaScript window.location redirect in the 404.html file the same as a 301 redirect for its indexing. From my testing I can confirm that Google will index all pages without issue, the only caveat is that the redirect query is what Google indexes as the url. For example, the url example.tld/about will get indexed as example.tld/?p=/about. When the user clicks on the search result, the url will change back to example.tld/about once the site loads.

closes #388, #105 
Will enable us to start working on: #258

### DIFF
--- a/DEVELOPING_TUTORIALS.md
+++ b/DEVELOPING_TUTORIALS.md
@@ -92,7 +92,7 @@ If proofing a PR for someone else, check out their branch (you'll see its name l
 > npm run serve
 ```
 
-6. Open a web browser to the following address to preview your work: http://localhost:3000/#/
+6. Open a web browser to the following address to preview your work: http://localhost:3000/
 
 Vue will update your localhost preview automatically as you make changes. However, you won't be able to see any newly added lessons until you've updated the appropriate routes and import statements, as described below.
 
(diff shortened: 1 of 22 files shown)


zebateira:
Let's revisit this later when we can say for sure all the implications of this are tackled.
For now the main issue to fix is the 404 Github page.

Issue #371, opened 2020-02-12
Maintain project references on smaller screens


terichadbourne:
We're currently using both text and a logo in our full-width version to call out the project a tutorial is about: 

![image](https://user-images.githubusercontent.com/19171465/74304983-9d3ad600-4d2c-11ea-8d63-5582c470d7e2.png)

However, when we go to mobile we remove that as though it's superfluous info: 
![image](https://user-images.githubusercontent.com/19171465/74305050-c52a3980-4d2c-11ea-97ce-bfc1dd95f477.png)

Only some of our tutorials have the project built into their name, which addresses the problem but is probably not the most ideal solution: 
![image](https://user-images.githubusercontent.com/19171465/74305084-e12ddb00-4d2c-11ea-992d-d3a6f1cd4059.png)

Note that Agata recently asked why we were duplicating the project name in these cases - it only feels like duplicating in the wider version: 
![image](https://user-images.githubusercontent.com/19171465/74305106-f3a81480-4d2c-11ea-9914-aaacbb8e80a0.png)

Let's pick a consistent naming/labeling system that doesn't lose any important info on mobile! (Not at all opposed to it being expressed in a different way in different sizes, but should be consistent across tutorials.)

terichadbourne:
closed by #396

PR #396, opened 2020-02-27
Feature: support multiple projects

zebateira:
This PR adds support for multiple projects by adding a new `projects.json` file. Worked on top of #329 PR. 
Current projects supported:

 - IPFS
 - IPLD
 - libp2p
 - Multiformats


zebateira:
The current issue with the long project name is:


image

terichadbourne:
The functionality is working great for me but I agree it won't work visually like this. :) 

Of the two options you suggested, I prefer Option 1 because it presents the project name in writing and doesn't rely on logo recognition. 
image
It also has the benefit of working naturally on mobile, which would close #371. 

A third option would be to use Option 1 on smaller screens, but pull the logo out bigger to the left on bigger screens (as it is now and in Option 2), leaving just the project title as shown above.

zebateira:
> A third option would be to use Option 1 on smaller screens, but pull the logo out bigger to the left on bigger screens (as it is now and in Option 2), leaving just the project title as shown above.

Here you go

## Option 3

Branch: https://github.com/ProtoSchool/protoschool.github.io/tree/feat/support-multiple-projects-option-3

### Desktop

<div>


</div>

### Small devices

<div>


</div>

zebateira:
IMO I prefer Option 3 if it wasn't for the disconnect between the logo and the project name. Having said that, I prefer Option 2.
Which one do you prefer @terichadbourne ?

terichadbourne:
Option 2 is a no go for me because the project title isn't spelled out anywhere. My first choice is Option 3 (because the logo placement in the big version feels a little less cluttered to me, although I take your point about the logo being separated from the project title) and second choice his Option 1.

zebateira:
@terichadbourne oh sorry, typo on my part. My preference is option 1, not 2.

mishmosh:
Option 3 makes the most sense to me, but perhaps with the project name _above_ the lesson name to preserve hierarchy.

terichadbourne:
Based on feedback from some design peers, we'll plan to go with Option 3 with the following modifications: 

- Set the project name in the site font (Avenir), not in the logotype. The symbol serves as the logo, in this case, so we don’t need two logos.
- Run the project name above the tutorial title, as a running head, essentially. It’ll look better, and it also makes more sense from an IA standpoint (eg Anatomy of a CID is a Multiformats tutorial, so Multiformats goes first).

terichadbourne:
Latest commit moves project name above tutorial and adjusts spacing accordingly. 

Fullscreen:

image

Mobile:

PR #397, opened 2020-02-27
fix: redirect any github 404 to our custom 404 page

### LINKED ISSUE #388: Bug: Loading GitHub default 404

zebateira:
Closes #388 

This should fix the 404 gibhub page showing.
@terichadbourne what we discussed regarding redirecting valid urls to hash urls would be a bit more tricky (e.g. `/tutorials` -> `/#/tutorials`) since it would still rely on the 404 hack #393

### DIFF
--- a/public/404.html
+++ b/public/404.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="utf-8" />
+
+  <title>ProtoSchool</title>
+
+  <script type="text/javascript">
+      window.location.replace('https://proto.school/#/404')
+    </script>
+</head>
+
+<body>
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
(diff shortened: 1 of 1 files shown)


terichadbourne:
@zebateira To attempt to proof this one I used: 
```sh
npm run build
npm install -g serve
serve -s dist
```

I got it running from dist but http://localhost:5000/404 redirects to http://localhost:5000/404#/ which is the homepage, not http://localhost:5000/#/404, our custom 404 page. I’m happy to assume this is a weird testing limitation and just merge and check when it’s live if you’re confident it’s set up as needed. Let me know what you recommend.

zebateira:
@terichadbourne weird behaviour from `serve`. I used this and it worked as it should:

`npm run build && npx http-server dist -p 3000`

PR #401, opened 2020-03-04
fix(content): clarify sample CID in  anatomy-of-a-cid

lidel:
I really enjoyed this tutorial, thank you for creating it! :heart: 

This PR applies a small correction to `tutorials/0006-anatomy-of-a-cid`.

The key takeaway here is to be explicit that the sample  CID is of the entire Wikipedia snapshot, not just that one sub-article. In the old version reader could think it is a CID of the single article.

Feel free to rewrite it if my English sounds wonky :-)

ps. If Wikipedia example is too complex, we could replace it with a standalone picture of aardvarks, and use its direct CID.


terichadbourne:
Thanks for calling this out, @lidel. I did assume it was just the aardvark page that the CID applied to, and I think getting into what I assume is a bigger issue of how websites are stored on IPFS (or what snapshots are) in this context may be too confusing to be worth it. Definitely open to suggestions if there's a fun photo or something else simple that would be a fun reference and also uses a CIDv0 so that we could do the same CIDv0 vs CIDv1 comparison at the end.

zebateira:
@terichadbourne I uploaded the photo in the wiki page to IPFS: https://ipfs.io/ipfs/QmcRD4wkPPi6dig81r5sLj9Zm1gDCL4zgpEj9CfuRrGbzF
Just need to pin it and we can use this example instead.

zebateira:
@lidel thank you for the input 🙏 
We'll take your second suggestion

>  If Wikipedia example is too complex, we could replace it with a standalone picture of aardvarks, and use its direct CID.

The wikipedia example is indeed a bit complex and using an image is simpler. Tackled in #402.

PR #408, opened 2020-03-17
feat: show message on tutorial if it has an update

zebateira:
## Update message

Screen Shot 2020-03-17 at 11 31 05 AM

## New tutorial message

Screen Shot 2020-03-17 at 12 59 30 PM

Message is optional

Screen Shot 2020-03-17 at 11 30 59 AM

Some notes:

 - The messages disappear after either:
    - the tutorial is completed after the tutorial update date
    - if it is a "new" message, it will disappear after one month of the publishing date of the tutorial
- The message is customizable in the property `updateMessage` in `tutorials.json`

### TODO:

- [x] If user has touched the tutorial, also show the update message
    - [x] change string 'passed' to date as well
    - [x] if any of the dates of the lessons are before the `updatedAt` date, then show the message
- [x] Docs
- [x] Tests


zebateira:
@terichadbourne I think I've bumped into an issue that is preventing easy testing for this PR.

For us to test this feature, we need to use mock data (a custom set of tutorials), but for that we need to jump between different sets of data: we want to have the full production data and a specific set for testing this feature (and possibly others in the future).

We are requiring the data through `require('tutorials.json')` and so switching between different json files is not trivial.

The first solution that comes to mind is to use a service worker to mock a server api. It would be used in production and then while testing we can simply mock any requests.
This solution however will require a bit of changes to replace all `require('tutorials.json')` with api calls and moving all the logic from the `utils/tutorials.json` to the "server". I would consider it a separate PR.

I'll continue to brainstorm for ideas to see if I can come up with something else easier that matches our setup.

zebateira:
I think I've managed to find an easier solution: set a window variable from cypress and then the `utils/tutorials.js` will load mock data from that variable. For this we need to use the option `onBeforeLoad` of `cy.visit()` like so:

```js
cy.visit(`/#/tutorials`, {
    onBeforeLoad (window) {
        window.__DATA__ = { tutorials }
    }
})
```

and then on `utils/tutorials.js`:

```js
import tutorials from '../static/tutorials.json'

const tutorialsData = window.__DATA__ || tutorials
```

terichadbourne:
Sounds like a great workaround for Cypress, @zebateira. I was just going to ask how I should go about testing this PR manually. Can I mess with my localStorage directly in the inspector in Chrome to create the conditions that will make a message be displayed or hidden?

zebateira:
@terichadbourne yeah if you want to force some conditions just mess around with localstorage to get the results you need.

src/utils/tutorials.js
```diff
@@ -91,11 +113,59 @@ export function getTutorialFullUrl (tutorialId) {
...
+    return lessonPassedAt && (
+      lessonPassedAt === 'passed' ||
+      new Date(lessonPassedAt) < updatedAtDate
+    )
+  })
+
+  return tutorialHasUpdates && (tutorialsPassedBeforeUpdate || someLessonPassedBeforeUpdate)
+}
+
+export function isTutorialNew (tutorial) {
```
terichadbourne, line 165: should show updated if any lessons have been completed since publishing

src/utils/tutorials.js
```diff
@@ -91,11 +113,59 @@ export function getTutorialFullUrl (tutorialId) {
...
 }
 
+export function hasTutorialBeenUpdatedRecently (tutorial) {
+  const updatedAtDate = new Date(tutorial.updatedAt)
+  const tutorialPassedAt = localStorage[`passed/${tutorial.url}`]
+  const tutorialHasUpdates =
+    tutorial.updatedAt !== tutorial.createdAt ||
+    !!tutorial.updateMessage
+
+  const tutorialsPassedBeforeUpdate = tutorialPassedAt && (
```
terichadbourne, line 148:
```suggestion
  const tutorialPassedBeforeUpdate = tutorialPassedAt && (
```

src/utils/tutorials.js
```diff
@@ -91,11 +113,59 @@ export function getTutorialFullUrl (tutorialId) {
...
+  const someLessonPassedBeforeUpdate = tutorial.lessons.some(lesson => {
+    const lessonPassedAt = localStorage[`passed/${tutorial.url}/${lesson.formattedId}`]
+
+    return lessonPassedAt && (
+      lessonPassedAt === 'passed' ||
+      new Date(lessonPassedAt) < updatedAtDate
+    )
+  })
+
+  return tutorialHasUpdates && (tutorialsPassedBeforeUpdate || someLessonPassedBeforeUpdate)
```
terichadbourne, line 162:
```suggestion
  return tutorialHasUpdates && (tutorialPassedBeforeUpdate || someLessonPassedBeforeUpdate)
```

src/utils/tutorials.js
```diff
@@ -117,8 +195,16 @@ export function getLessonType (tutorialId, lessonId) {
...
 }
 
-export function getTutorialByUrl (tutorialUrl) {
-  return Object.values(tutorials).find(({ url }) => url === tutorialUrl)
+export function setLessonPassed (tutorial, lesson) {
+  if (lesson.type === 'resources') {
+    localStorage[`passed/${tutorial.url}/resources`] = new Date().toISOString()
+  }
+
+  localStorage[`passed/${tutorial.url}/${lesson.formattedId}`] = new Date().toISOString()
```
terichadbourne, line 203:
```suggestion
  if (lesson.type === 'resources') {
    localStorage[`passed/${tutorial.url}/resources`] = new Date().toISOString()
  } else {
     localStorage[`passed/${tutorial.url}/${lesson.formattedId}`] = new Date().toISOString()
```

cypress/integration/tutorial-messages.spec.js
```diff
@@ -94,6 +94,24 @@ describe(`DISPLAY TUTORIAL MESSAGES`, function () {
       cy.get('.tutorial-message--new .tutorial-message-text').should('contain', 'new message')
     })
 
+    it(`should not show new message for tutorials published less than a month ago and the user has some lessons completed`, function () {
```
terichadbourne, line 97:
```suggestion
    it(`should not show new message for tutorials published less than a month ago in which the user has completed some lessons`, function () {
```

zebateira:
@terichadbourne implemented the message we talked about.
To trigger it, you need to:

- change the code to fail (previous working code for example)
- have completed the lesson before (you can set it to `passed` in the `localStorage` directly)
- refresh the page

If it's a code lesson only, the error and the message will show. If it's a file upload message, then you'll see the message to add the cached files again, and then after adding the files you'll see the message.

terichadbourne:
@zebateira Looking good! This is showing at the appropriate times. I pushed a commit to tweak the wording slightly. If you like it and tests are passing, please go ahead and merge!

PR #415, opened 2020-04-02
Support virtual event listings

### LINKED ISSUE #409: Add support for virtual event listings
A few tweaks we can make to support community leaders who'd like to explore virtual events: 

- [ ] Tweak form to allow "Virtual" in place of location
- [ ] Add timezone support, since location will no longer be an indicator of time zone

terichadbourne:
**WIP** - Closes #409 (once complete)

Changes to date: If an event record in `events.json` includes a `virtual` property set to the string `true`, the resulting card displays with "VIRTUAL" in place of "CITY" and does not show region or country. 

image

Based on the current logic, only the String `true` works for this. A Boolean `true`, a missing `virtual` property, or a String /Boolean `false` will all lead to a standard live event card with location displayed. 

Note that the time provided by the organizer is now useless because we don't have a location as a clue to what timezone it applies to. 

**To do in form:** 
- [x] Add virtual checkbox
- [x] Add timezone support 
- [x] Make city and country optional (either generally or when virtual checkbox is checked - the latter would require splitting the form into sections)

**To do in code:** 
- [x] Update data fetch to add `"virtual": "true",` in the events.json where appropriate
- [x] Update data fetch to pull in timezones (submitted in form in timezone of event organizer)
- [x] Update UI to display timezones (probably most useful if it displays in the user's time zone, with time zone clearly spelled out)


terichadbourne:
@zebateira This PR is currently set to Option 3, which I think is my favorite out of the options below. Let me know what you like the looks of. 

Option 1 
image

Option 2
image

Option 3
image

Option 4
image

I've gone through the form and added the UTC offsets as of the event dates so that all records will have time zones to display. 

I think this is in a state where we can merge it (after removing the commented-out code for the alternate display options), but our form has a significant bug in that it's not dealing with daylight savings properly.  My recommendation is that we merge this and then open a new issue for a better timezone plan.

terichadbourne:
@zebateira This PR used to have failing Travis builds on it, and then I fixed a typo and merged the latest updates to get tests passing locally, pushed everything, and now the Travis stuff is entirely missing. Travis status claims all systems are operational, so no clue what happened there. ¯\_(ツ)_/¯

zebateira:
@terichadbourne thinking through our bug, actually, I don't think we have any bug.
If I create an event in August 29th at 3pm, that's the time it will happen, regardless of the daylight saving time. So I think the current implementation is ok.
The only weird thing that I can spot is that the current GMT offsets that I provided to the Google Forms has the current offsets based on the current daylight savings time. Meaning in 6 months, some GMT offsets might be different, but I think it's acceptable since the city name is present, so I don't see it as a big issue. What do you think?

PR #418, opened 2020-04-09
Toggle code tutorials tracking & caching

### LINKED ISSUE #366: Feature: Cache and track toggle status for hiding coding tutorials
Since introducing our lesson/tutorial type icons and adding the toggle button to hide coding tutorials (#346), we now have the opportunity to cache the toggle state so that if a user has previously chosen to hide coding tutorials they'd find them hidden when returning to the tutorials page. (Currently it loads/reloads with all tutorials visible.)

An additional bonus would be if we created a parameter in the URL that would set the toggle to hide the coding tutorials, so that we could give someone a URL to go to to see just the non-coding ones.

zebateira:
This PRs adds tracking and caching to the code tutorials toggle in the `/tutorials` page after @terichadbourne started this effort initially!

## Tracking

The following data is sent to countly when the toggle is turned off:

```js
{
    key: 'filter',
    segmentation: {
        path: '/tutorials',
        filteredData: 'tutorials',
        filter: 'hideCodingTutorials',
        method: 'toggle'/'urlQuery'
    }
}
```

The two methods to track:

-  `toggle`: when the toggle is clicked to turn it off
- `urlQuery`: when the toggle is turned off based on the query parameter `?code=false`

## Caching

The value of the toggle is cached in `localStorage` in the new `settings` namespace.
The localstorage key is `settings/filters/tutorials-show-coding` for this specific filter.
See the docs for the new `settings` module to understand how to use it.

### TODO

 - [x] Tests

Closes #366


terichadbourne:
@zebateira The new toggle caching and URL query appear to be working correctly. However, the vast majority of our tests weren't running anymore, so I pushed a commit that removes `.only` from the `describe` you were focused on testing. Feel free to merge if you agree this fixes that problem.

PR #429, opened 2020-04-24
feat: add newsletter subscription form component

### LINKED ISSUE #422: Subscribe Newsletter box and News page
New subscribe newsletter box for users to subscribe to ProtoSchool Newsletter.

image

Figma: https://www.figma.com/file/lnCeFi9O1PIgxMdKcgHIcM/ProtoSchool-Website?node-id=93%3A1699

- [ ] Subscribe Newsletter UI
- [ ] Mailchimp connection
- [ ] After user has subscribed successfully, cache the result so we can hide the subscription box in the future
- [ ] Add it to Resources page
- [ ] Add to new /news page
- [ ] Track to Countly
- [ ] Final copy

zebateira:
This PR adds a Newsletter Subscription form integration with Mailchimp. Closes #422 

Since mailchimp integration is pending (see issue bellow), I've made the form succeed after 1.5 seconds.

# TODO

- [x] Mailchimp Integration
    - Mailchimp does not support client-side implementations (no CORS support) so this means we need a proxy server to finish the integration
    - Another alternative would be to go back to the idea of customising an embedded form - this would not allow for the flexible customisation that we have right now.
- [x] Add /news page and include the newsletter subscription form as well.
- [x] Accessibility testing
- [x] Cache form submitted to hide it after
- [x] Final copy

UPDATE: I've managed to make the mailchimp integration work. The only thing left to do now is to add some variables to 1Password and travis and we're good to go.


terichadbourne:
@zebateira I just realized that I hadn't added the necessary paragraph formatting to make this page match the others, so I've now made that update. 

One little bug I haven't yet diagnosed:  When I click on any other nav item, the location of the content stays consistent. When I click on the new News tab, it looks like all the content moves a smidge to the right. Any idea why that might be happening?

zebateira:
@terichadbourne 

> One little bug I haven't yet diagnosed: When I click on any other nav item, the location of the content stays consistent. When I click on the new News tab, it looks like all the content moves a smidge to the right. Any idea why that might be happening?

Ah that's happening because this page on big screens does not need a scrollbar. If you resize the window to reduce the height you'll see the scrollbar being added and the that behaviour does not happen anymore. I pushed the fix for this:

```css
html {
  overflow-y: scroll;
}
```

PR #438, opened 2020-05-08
fix: use ipfs error codes as much as we can to translate error messages

### LINKED ISSUE #343: Improve strategy for validating code exercises by using Error Codes instead of relying on Error Messages from IPFS lib
To continue the previous discussion on #337, I think it is worth considering that we make validation code use the error codes instead of the error messages so that our logic becomes more robust.

If we go forward with this, it needs to be standardized and documented, but at the same time, we could apply this new approach iteratively on new tutorials only, and when possible we could update old ones and add tests for them as well so we are more confident that we are not breaking anything.

zebateira:
I tested the cases I changed. Let me know if you can spot anymore or if you have any questions.

Closes #343


src/tutorials/0004-mutable-file-system/08.js
```diff
@@ -62,9 +62,9 @@ const validate = async (result, ipfs) => {
     }
   } else if (rootIsEmpty) {
     return { fail: 'Your root directory is empty. Did you accidentally move the `some/stuff` directory? Remember to test whether each item is a file (`type === 0`) before moving it.' }
-  } else if (result instanceof Error && result.message === 'paths must start with a leading /') {
+  } else if (result instanceof Error && result.code === utils.ipfs.errorCodes.ERR_INVALID_PATH) {
     return {
-      fail: 'Paths must start with a leading `/`. Did you use just the file name when attempting to move each file?',
+      fail: 'Invalid path. Did you use just the file name when attempting to move each file? Remember to start the path with a leading `/`',
```
terichadbourne, line 67:
```suggestion
      fail: 'Invalid path. Did you use just the file name when attempting to move each file? Remember to start the path with a leading `/`.',
```

src/tutorials/0004-mutable-file-system/09.js
```diff
@@ -13,14 +13,15 @@ const validate = async (result, ipfs) => {
 
   try {
     someStuffFiles = await all(ipfs.files.ls('/some/stuff'))
-  } catch (err) {
-    if (err.code === 'ERR_NOT_FOUND') {
+  } catch (error) {
+    if (error.code === utils.ipfs.errorCodes.ERR_NOT_FOUND) {
       return {
-        fail: 'The directory `/some/stuff` did not exist - try resetting the code and starting again'
+        fail: 'The directory `/some/stuff` does not exist - try resetting the code and starting again',
```
terichadbourne, line 19:
```suggestion
        fail: 'The directory `/some/stuff` doesn't exist. Try resetting the code and starting again.',
```

src/tutorials/0005-regular-files-api/07.js
```diff
@@ -7,23 +7,26 @@ const validate = async (result, ipfs) => {
     }
   }
 
-  if (result.error) {
-    if (result.error.toString().includes('file does not exist')) {
+  if (result instanceof Error) {
+    if (result.code === utils.ipfs.errorCodes.ERR_NOT_FOUND) {
       return {
-        fail: "Oops, we could not find a file with that IPFS path. Are you sure you're using the correct path with the correct CID? Remember, if you use the wrapping directory's CID, you need to append `/fun/success.txt` to the path name. Otherwise, if you're using the `fun` subdirectory's CID', you need to append `/success.txt`."
+        fail: "Oops, we could not find a file with that IPFS path. Are you sure you're using the correct path with the correct CID? Remember, if you use the wrapping directory's CID, you need to append `/fun/success.txt` to the path name. Otherwise, if you're using the `fun` subdirectory's CID', you need to append `/success.txt`.",
```
terichadbourne, line 13:
```suggestion
        fail: "Oops, we could not find a file with that IPFS path. Are you sure you're using the correct path with the correct CID? Try checking for typos. Remember, if you use the wrapping directory's CID, you need to append `/fun/success.txt` to the path name. Otherwise, if you're using the `fun` subdirectory's CID, you need to append `/success.txt`.",
```

src/tutorials/0004-mutable-file-system/08.js
terichadbourne, line 67:
```suggestion
      fail: 'Invalid path. Did you use just the file name when attempting to move each file? Remember to start the path with a leading `/`.',
```

src/tutorials/0005-regular-files-api/07.js
```diff
@@ -7,23 +7,26 @@ const validate = async (result, ipfs) => {
...
+        overrideError: true
       }
     }
-    if (result.error.toString().includes('this dag node is a directory')) {
+    if (result.toString().includes('this dag node is a directory')) {
       return {
-        fail: 'The `cat` method only works on files, but you tried to use it on a directory. Did you forget to include the relative file path?'
+        fail: 'The `cat` method only works on files, but you tried to use it on a directory. Did you forget to include the relative file path?',
```
terichadbourne, line 19:
```suggestion
        fail: 'The `cat` method only works on files, but you tried to use it on a directory. Did you forget to include the relative file path? Remember, if you use the wrapping directory's CID, you need to append `/fun/success.txt` to the path name. Otherwise, if you're using the `fun` subdirectory's CID, you need to append `/success.txt`.',
```

PR #443, opened 2020-05-29
fix: update ipfs docs links

### LINKED ISSUE #339: Update IPFS docs links when legacy site is deprecated
Once the content of https://docs.ipfs.io/ is replaced by the content of https://docs-beta.ipfs.io/, we should search our codebase for links to docs.ipfs.io and check that they're all going to valid entries, updating URLs if needed. (Much of the content has been regorganized and will have new URLs.) 

I've added a reminder in the [docs deprecation plan](https://hackmd.io/-JU3CcXgSgmcXWszyku2PA) for us to circle back to this when the time comes.

zebateira:
This PR changes the ipfs docs site links to the new URLs after the update.
Redirects have been setup, so this is not a fix, just an improvement so users skip the redirects.

Closes #339


terichadbourne:
Looks great. Thanks for taking the time to do this!

PR #444, opened 2020-05-29
fix: open mailchimp submission in new tab to firefox issue with content blocking

### LINKED ISSUE #441: Bug: Newsletter subscription blocked on Firefox
Newsletter signup apparently blocked on Firefox. See https://twitter.com/raybachas/status/1263507550135296002?s=21

@zebateira Mind taking a look after the weekend to see if you can reproduce and diagnose?

zebateira:
This approaches disables the jsonp request and uses the good old form action method.

success-480p

This fixes the issue in firefox issue when content blocking is enabled (closes #441).
The downside of this is that we can't know if the subscription fails, and so now we are always showing the success message which might not match what the user has seen in the mailchimp browser tab if an error occurs. For example:

error-480p

Maybe we can just change the copy to a more generic message? This way it won't be so weird in these error edge cases.

Let me know what you think @terichadbourne


zebateira:
image

This is how the message looks like now

PR #449, opened 2020-06-12
Create DWeb Basics pseudo-project

terichadbourne:
For tutorials that are about general DWeb concepts rather than a specific project, this PR creates a "DWeb Basics" project label that currently uses the ProtoSchool logo. I'm very open to suggestions for either renaming or re-logoing with some other generic icon. 

Some name options: 
DWeb Basics
DWeb Primer
DWeb Concepts

image
image


jessicaschilling:
+1 for "Dweb Concepts" (double-check with @johnnymatthews about whether the W is capitalized, I don't remember!) ... not all of them are truly basic, and this makes it clear that you're talking about larger-spanning overall concepts, foundational building blocks, etc.

mishmosh:
My 2 cents: I like DWeb Concepts.

terichadbourne:
I've changed to DWeb Concepts based on feedback (but open to more). 

> "Dweb Concepts" (double-check with @johnnymatthews about whether the W is capitalized, I don't remember!)

If you're not familiar with Decentralized Web, there would be no reason to know the d was an acronymn in Dweb (just reads like a nonsense word or a mispelling of dweeb). I checked and the recent Fluence survey uses the capital W.


@zebateira Do you remember where these URLs are being used? Tried searching the codebase for project.url and didn't find anything so it may just be future-proofing. 
image

zebateira:
@terichadbourne yeah the url is for future use cases. Not used ATM 👍

PR #456, opened 2020-06-23
docs: fix typos

pgrimaud:
Hey,

This PR will fix : 

- 1 typo in `CODE_OF_CONDUCT.md`
- 1 typo in `DESIGNING_TUTORIALS.md`
- 2 typos in `DEVELOPING_TUTORIALS.md`
- 1 typo in `src/tutorials/0004-mutable-file-system/09.md`
- 1 typo in `src/tutorials/0004-mutable-file-system/10.md`
- 1 typo in `src/tutorials/0006-anatomy-of-a-cid/02.md`

Bye! :robot:

PR #461, opened 2020-06-25
Fix: handle PRs from forks failing builds

zebateira:
If a build from a PR that is started from a forked project starts, travis does not allow the secure env vars to be used. See here: 

- Travis issues: [travis#701287172](https://travis-ci.org/github/ProtoSchool/protoschool.github.io/builds/701287172) [travis#701462759](https://travis-ci.org/github/ProtoSchool/protoschool.github.io/builds/701462759)
- this is because secure env vars are not available outside the repo
    - [travis-ci#1946](https://github.com/travis-ci/travis-ci/issues/1946)
    - [travis-ci#6611](https://github.com/travis-ci/travis-ci/issues/6611)

Documented here: [https://docs.travis-ci.com](https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml)

> Encrypted environment variables are not available to pull requests from forks due to the security risk of exposing such information to unknown code.

image

We need to still run the build process, but skip the data build scripts.
For this we added a env var to travis `BRANCH=code` that is only set in the `code` branch.

So the logic is as follows:

1. If no config was found
    - And no `BRANCH=code` env var is defined, we ignore the error and exit with code 0
    - If `BRANCH=code` env var is defined, we fail the script with exit code 1
2. If config was found, proceed normally (it's the cron job or a website publish)

Fixes #456 and #459 (will rebase them into `code` after we get this merged)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+os:
+  - linux
+dist: xenial
 language: node_js
 node_js:
   - lts/*
@@ -26,11 +29,11 @@ notifications:
   email: false
 deploy:
   provider: pages
-  local-dir: dist
-  target-branch: master
-  skip-cleanup: true
-  github-token: $GITHUB_TOKEN  # Set in the settings page of your repository, as a secure variable
-  keep-history: true
+  local_dir: dist
+  target_branch: master
+  cleanup: true
+  token: $GITHUB_TOKEN  # Set in the settings page of your repository, as a secure variable
+  keep_history: true
   fqdn: proto.school
   on:
     branch: code
--- a/package-lock.json
+++ b/package-lock.json
@@ -7138,9 +7138,9 @@
       "dev": true
     },
     "err-code": {
-      "version": "2.0.0",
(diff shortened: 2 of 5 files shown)


zebateira:
To test this PR, remove (or temporarily rename) the file `.env` so that we simulate the situation of no available config. Then, run two commands:

1. `npm run scripts:build:data` should exit with code `0` (to check the exit code run `echo $?` right after the previous command)
2. `BRANCH=code npm run scripts:build:data` which should exit with code `1`.

This happens because only travis will add the `BRANCH=code` environment variable.

terichadbourne:
@zebateira I believe this is working as intended, but since I don't explicitly see exit codes I'll paste my results so you can confirm. If these screenshots look correct to you, please feel free to merge. Thanks!

_**Having renamed my .env file and running this locally:**_ 

image


**_Having restored my local .env file:_** 

image

zebateira:
@terichadbourne yeah that's it 👌

Issue #454, opened 2020-06-22
Lesson Feedback%3A Verifying Storage on Filecoin - Lesson 3 (Proof-of-Replication (PoRep))


porcuquine:
Have a question or suggestion regarding a specific ProtoSchool lesson? Please use this template to share it!

**URL of the lesson that's confusing:**

https://proto.school/#/verifying-storage-on-filecoin/03

**What's confusing about this lesson?**

> During sealing, the sector data (identified by the CommD) is encoded through a sequence of graph and hashing processes to create a unique replica. The root hash of the resulting replica is the CommR (Commitment of Replication, aka SealedSectorCID), which is recorded to the blockchain.

This is not quite accurate — or, the way in which it is accurate is complicated. I would not describe `CommR` as the root hash of the replica. It is a commitment to the replica, though. If just changing that wording works for you, it works for me. Otherwise, a more elaborate explanation may be required if you want to be accurate in a way that's also not potentially misleading.

**What additional context could we provide to help you succeed?**

**What other feedback would you like to share about ProtoSchool?**

terichadbourne:
Thanks for the feedback @porcuquine! @mishmosh what do you recommend for beginner-friendly rewording here?

As a novice, viewing a CID as "a commitment to a replica" doesn't mean much to me. As an example of the vocab in normal English, I make a commitment to my landlord to pay rent, and the proof that I made that commitment is my lease. Are the CommD and CommR vaguely lease-like in this analogy? Or are they a CID (hash) of the lease? Or neither? 😂

porcuquine:
Regarding normal English vocab: 'commitment' in this sense is not like 'obligation'. It's more like writing a prediction in a sealed envelope and putting the sealed envelope in a public place. Later, when I claim to have predicted the eventual outcome, I open the envelope and show that I had indeed done so. You can trust that it was a real prediction because I committed to it in advance. If I only produce the envelope after the event has occurred, my uncommitted prediction doesn't mean much. As an aside, I would try to distinguish technical terms (in this context) from normal English vocab or things are going to get confusing.

I'm not going to address 'lease' in this context because it either doesn't apply or I don't understand the translation you intend well enough.

CommD *is* the root hash of a merkle tree (of the original data). So if that makes sense to you and your readers, it's fine.

There exists a parallel value, CommRLast — which is also the root hash of a merkle tree (of the replica). If you were talking about CommRLast, where you mention CommR, everything would still make perfect sense.

[...]

PR #464, opened 2020-06-26
Address inaccuracies in Verifying Storage on Filecoin 

### LINKED ISSUE #455: Lesson Feedback%3A Verifying Storage on Filecoin - Lesson 4 (Proof-of-Spacetime (PoSt))
Have a question or suggestion regarding a specific ProtoSchool lesson? Please use this template to share it!

**URL of the lesson that's confusing:**

https://github.com/ProtoSchool/protoschool.github.io/issues/454

**What's confusing about this lesson?**

>Rather than checking for the original data, it checks the encoded version (the CommR), which during the Proof of Replication was proved to be an encoding of the original data. Since the CommR is a type of merkle tree called a Directed Acyclic Graph (DAG), this can be done with a merkle inclusion proof.

Some adjacent but independent points:
- `CommR` is not the root of the merkle tree built over the replica.
- The root of that merkle tree is not the tree itself.
- The reference to DAGs is out of place.

I tried to just rewrite the above to say what I think you mean, but how to do that will somewhat depend on changes to the previous explanation also, unfortunately. So let me at least explain what you need to make a fuller explanation, and you can choose the right level of granularity and use it consistently.

- `CommR` is a commitment to the replica, but it is a hash of a tree used only in PoRep *and* of the commitment we call `CommRLast` — which doesn't appear on chain. `CommRLast` is indeed the merkle root of the replica (i.e. the uniquely encoded copy of the original data).
- Since `CommRLast` is a merkle root — the result of having constructed a merkle tree over the replica, proof of each random check can be provided with a merkle incusion proof.

I think you may want to elide some of the details of how `CommR` is constructed and used, but please try to avoid doing so in a way suggesting that `CommR` is actually the merkle root of the replica. This is not true and could confuse someone seeking a deeper understanding.

**What additional context could we provide to help you succeed?**

**What other feedback would you like to share about ProtoSchool?**

terichadbourne:
Closes #455 #454 

This PR aims to address the inaccuracies pointed out by @porcuquine in the Verifying Storage on Filecoin tutorial. Look forward to confirmation that this attempt at rephrasing is actually true while still being beginner-friendly. 

co-authored by @mishmosh

### DIFF
--- a/src/tutorials/0007-verifying-storage-on-filecoin/03.md
+++ b/src/tutorials/0007-verifying-storage-on-filecoin/03.md
@@ -15,7 +15,9 @@ Once a sector is full, a CommD (Commitment of Data, aka UnsealedSectorCID) is pr
 
 Next, a process called _sealing_ takes place.
 
-During sealing, the sector data (identified by the CommD) is encoded through a sequence of graph and hashing processes to create a unique replica. The root hash of the resulting replica is the CommR (Commitment of Replication, aka SealedSectorCID), which is recorded to the blockchain.
+During sealing, the sector data (identified by the CommD) is encoded through a sequence of graph and hashing processes to create a unique replica. The root hash of the merkle tree of the resulting replica is the CommRLast. (Learn more about merkle trees in our [Decentralized Data Structures](https://proto.school/#/data-structures) tutorial.)
+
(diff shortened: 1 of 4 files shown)


terichadbourne:
Thanks so much for spending the time on feedback and review @porcuquine! Much appreciated.

PR #472, opened 2020-07-10
Fix: do not track multiple times lessonPassed and tutorialPassed events to Countly

### LINKED ISSUE #470: Confirm we're not double-logging tutorialPassed or lessonPassed events
Countly shows that users are passing each tutorial more than once per visitor: 

image

Let's revisit our tracking code with fresh eyes to confirm we're tracking both`lessonPassed` and `tutorialPassed` at appropriate times and avoiding sending repeat values from the same visitor.

zebateira:
This PR fixes the issue of sending multiple events to Countly, when they should have been only sent once.
The events that should only be sent once are `lessonPassed` and `tutorialPassed`. All other events will continue to be sent multiple times.

To fix this, I added a `trackEventOnce` method to a new countly util that uses `LocalStorage` (settings module) to keep a record of which events have been sent that should not be repeated.
To uniquely identify an event, we use a combination of the event name and the data sent. This way, we can send multiple events of `lessonPassed` for different lessons, but not for the same lesson. Examples:

  1. Tracking the same event with different data -> both events are tracked
  `event: 'lessonPassed', data: { lessonNumber: 1 }`
  `event: 'lessonPassed', data: { lessonNumber: 2 }`

  2. Tracking the same event with the same data -> only the first event is tracked
  `event: 'lessonPassed', data: { lessonNumber: 1 }`
  `event: 'lessonPassed', data: { lessonNumber: 1 }`

Internally, the event id (localstorage item key) would look something like this:

```
settings/countly/event-traked/lessonPassed-{"tutorial":"Data Structures","lessonNumber":1,"path":"/data-structures/01"}
```

This approach will simplify code logic of having to check if we should or not send an event to countly. We just need to use `countly.trackEvent` or `countly.trackEventOnce` depending on our use case.
I've changed the common strategy of tracking an event from using the method `trackEvent` in a component to a computed property `trackingData`. To send the event we just need to call `countly.trackEvent(event, this.trackingData)`, which allows to add or remove properties from the data to be sent to countly.

I've manually tested all events to ensure we are sending them correctly at the right times and that the data sent is correct, but a fresh pair of eyes doing the testing as well would be best (adding a `Countly.debug = true` to the countly config in the `index.html` file helps + filtering the console by "adding event lessonpassed" for example).

Closes #470


terichadbourne:
Thanks for this @zebateira! I've tested all of the events I see listed in the countly util. Other than the issues you already addressed, here's what I spotted that I'd recommend addressing, either here or in another PR: 

- [ ] `tracking` feels like an odd variable name to use for the source we send to tracking in the NewsletterSubscription component. The code would feel cleaner to me if we called it `source` or `trackingSource`, or as you mentioned it could be calculated within the component rather than passed. I definitely want to continue passing a source (resources or news) rather than other tracking path, so that we don't have to add up a bunch of different resources paths to compare to the news path. 

- [ ] It looks to me like the tracking for dismissing the survey is only being sent if you dismiss it after completing the survey (when the questions themselves are gone), and not when you dismiss it while the questions are still there as a way to decline doing it. I'd expect the reverse. 

- [ ] It looks like we're tracking the abtesting option A or B every time they reload the resources pages as opposed to once per resources page. I feel like the latter would make more sense so we can get to a stat of "of X people presented with the option to complete an Option A survey for tutorial Y, Z people completed it." I do think we should still count it multiple times per person, just only once per tutorial.

zebateira:
1. Tracking prop naming: [PR#474](https://github.com/ProtoSchool/protoschool.github.io/issues/474)
2. We are always tracking the event: when the user hasn't answered any questions, one more questions or when they finished it. We actually send 2 properties that indicate the progress of the user: `numberOfQuestionsAnswered` (0-3) and `surveyCompleted` (true of false)
3. I just pushed a fix for the AB Testing issue you mentioned.

terichadbourne:
Good to go after we merge #474 into this.

PR #489, opened 2020-07-17
Fix: tests failing on travis

zebateira:
This is a short term solution I found to make the tests pass in Travis: increase the timeouts.
The reason why some tests fail seems to be because submitting the code solution takes some time. up to 30 seconds sometimes, but it can spike further - 60s timeout was not enough in some rare occasions.
So it seems that some ipfs operations are taking longer than desirable. I'm not sure why this is the case only in travis. Locally (and production) it's not that slow, but I was able to reproduce it running on electron locally.

### DIFF
--- a/src/components/Lesson.vue
+++ b/src/components/Lesson.vue
@@ -112,6 +112,7 @@ import all from 'it-all'
 import toBuffer from 'it-to-buffer'
 import newGithubIssueUrl from 'new-github-issue-url'
 
+import { isProduction } from '../utils/env'
 import {
   getTutorialByUrl,
   isTutorialPassed,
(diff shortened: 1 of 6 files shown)


src/components/Lesson.vue
```diff
@@ -136,7 +137,7 @@ import Validator from './Validator.vue'
 import TutorialCompletionCallout from './callouts/TutorialCompletion.vue'
 import TypeIcon from './TypeIcon.vue'
 
-const MAX_EXEC_TIMEOUT = 10000
+const MAX_EXEC_TIMEOUT = isProduction ? 10e3 : 60e3
```
terichadbourne, line 140: @zebateira Can you teach me why these aren't numbers? Or is this some fancy number formatting I haven't learned yet?

terichadbourne: Figured it out but would suggest that these numbers are small enough that it makes sense to write them out to make the code easier for newcomers to follow.

zebateira, line 140:
```suggestion
const MAX_EXEC_TIMEOUT = isProduction ? 10000 : 60000
```

zebateira: @terichadbourne changed 👍

PR #488, opened 2020-07-17
fix: upgrades all tutorials to the latest js-ipfs 0.48.0

achingbrain:
Also fixes loads of broken links to the core api docs.

See the [release notes](https://github.com/ipfs/js-ipfs/releases/tag/ipfs%400.48.0) for 0.48.0 for API changes.

Notably `ipfs.add` now only returns single items so you can swerve the whole "what's an AsyncIterator then?" conversation.

`ipfs.addAll` now exists for when you want to add multiple items and returns an AsyncIterator, basically the old behaviour of `ipfs.add`.

PR #490, opened 2020-07-20
fix: add update messages to basics an mfs tutorials

zebateira:
Because of the changes from #488, we need to update the Basics and the MFS tutorial messages as well.

### DIFF
--- a/src/static/tutorials.json
+++ b/src/static/tutorials.json
@@ -79,9 +79,9 @@
     "title": "P2P Data Links with Content Addressing",
     "description": "Store, fetch, and create verifiable links between peer-hosted datasets using the IPFS DAG API and CIDs. It’s graphs with friends!",
     "newMessage": "",
-    "updateMessage": "",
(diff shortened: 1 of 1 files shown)


src/static/tutorials.json
```diff
@@ -79,9 +79,9 @@
     "title": "P2P Data Links with Content Addressing",
     "description": "Store, fetch, and create verifiable links between peer-hosted datasets using the IPFS DAG API and CIDs. It’s graphs with friends!",
     "newMessage": "",
-    "updateMessage": "",
+    "updateMessage": "Version `0.48` of `js-ipfs` has introduced some changes, most notably by adding `files.addAll` to use with multiple files, while `files.add` can now only be used with a single file. This means that **some code challenge solutions have changed**. We recommend re-visiting this tutorial to learn about the updated methods and complete the revised challenges!",
```
zebateira, line 82:
```suggestion
    "updateMessage": "Version `0.48` of `js-ipfs` has introduced some changes, most notably by changing the `ipfs.dag.get` method to use a structured object for the path of the object. This means that **some code challenge solutions have changed**. We recommend re-visiting this tutorial to learn about the updated methods and complete the revised challenges!",
```

PR #493, opened 2020-07-22
fix: lost references to the files.add method changed to files.addAll

zebateira:
The only affected tutorial is "Regular Files API". Affected lessons are lessons number 3, 5 and 6.

### DIFF
--- a/src/static/tutorials.json
+++ b/src/static/tutorials.json
@@ -212,7 +212,7 @@
     "title": "Regular Files API",
     "description": "The IPFS Regular Files API provides a way to store and share files in a peer-to-peer storage system.",
     "newMessage": "",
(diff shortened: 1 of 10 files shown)


zebateira:
From the [convo](https://protocollabs.slack.com/archives/CHPA2EMG8/p1595868442473600?thread_ts=1595613336.466400&cid=CHPA2EMG8) with @achingbrain 

> the plan for the future is to make it throw if you pass input that would result in multiple files being created

So I propose leaving the issue as it is: a user will be able to use `ipfs.add(files)`. In the future we will get a release that will make that specific case fail, and then we'll be able to show an appropriate error message to the user.

terichadbourne:
Merging as is for now, noting that we'll have to come back and update validation when the js-ipfs team settles on some more under-the-hood edits to how these methods work. See #498.

PR #505, opened 2020-07-30
Feature: improve SEO metadata by adding vue-meta

### LINKED ISSUE #258: Feature: Enable separate metadata for different SPA page views for social sharing

zebateira:
Closes #258

This PR adds [`vue-meta`](https://github.com/nuxt/vue-meta) and an util `head` to help adding all the metadata to all the pages, which helps add separate meta data tags information for each page and improve SEO.

There are more SEO tweaks that we can do further, but I'll leave that to another PR (will tackle #506).


terichadbourne:
@zebateira Can you please look over my last two commits, especially the one that programatically adds the project name to the tutorial landing page's title metadata if the tutorial title doesn't include it? 

Can you also help me understand how to override the template that adds `| ProtoSchool` to everything so that the homepage title can be changed to `ProtoSchool | Interactive tutorials on decentralized web protocols`?

PR #511, opened 2020-08-04
Fix: remove netlify cypress plugin and run cypress in the build command

zebateira:


zebateira:
Closing this since build is not working properly.

Issue #520, opened 2020-08-07
Fix: prerender redirect pages so redirects can be handled on client side


zebateira:
By using the [`postProcess` option](https://www.npmjs.com/package/prerender-spa-plugin#using-the-postprocess-option) we can prerender the old pages and then let the client redirect the user.

The only issue here would be SEO, so I'd still recommend redirects to be configured on server side so crawlers can receive the 301 code.
The main issue we need to fix is to not rely on our hosting about the redirects, but still try to setup the redirects there to get better SEO results.

zebateira:
Note that we also need to update protowizard and the intructions on the [developing guides tutorial](https://github.com/ProtoSchool/protoschool.github.io/blob/main/DEVELOPING_TUTORIALS.md#renaming-a-tutorial-after-it-has-been-published).

zebateira:
Updating the strategy considering the future move to fleek.

Issue #506, opened 2020-07-30
More SEO Improvements


zebateira:
We need to improve our SEO stats. To do this we can improve the following:

- [x] Only have one `h1` tag per page (ATM he have 2 in some pages)
- [x] Add `title` and `alt` to all images (missing in some)
- [x]  Let homepage have a different format for its title
- [x] Sitemap
    - Add trailing slashes to the sitemap paths
    - Add more properties to each sitemap entry: `priority`, `lastmod` and `changefreq`
- [x] Allow to override html title tag without the `| ProtoSchool` suffix

This is what's missing considering the changes in #505.

terichadbourne:
Other things not yet done at time of merging #505: 

- [ ] Let homepage have a different format for its title
- [ ] Set tutorial landing pages as canonical links for lesson pages

zebateira:
> Set tutorial landing pages as canonical links for lesson pages

Regarding this, I did some research and this is what I found:

https://moz.com/learn/seo/canonicalization:
> 5. Be careful canonicalizing near-duplicates
When most people think of canonicalization, they think of exact duplicates. It is possible to use the canonical tag on near-duplicates (pages with very similar content), but proceed with caution. There’s a lot of debate on this topic, but It’s generally ok to use canonical tags for very similar pages, such as a product page that only differs by currency, location, or some small product attribute. **Keep in mind that the non-canonical versions of that page may not be eligible for ranking, and if the pages are too different, search engines may ignore the tag.**

https://yoast.com/rel-canonical/: 
> Using rel=canonical on not so similar pages
While I wouldn’t recommend this, you can use rel=canonical very aggressively. Google honors it to an almost ridiculous extent, where you can canonicalize a very different piece of content to another piece of content. However, if Google does catch you doing this, it will stop trusting your site’s canonicals and thus cause you more harm…

https://www.gsqi.com/marketing-blog/google-ignore-rel-canonical-different-content/:
> Google’s Algorithms Can Ignore Rel Canonical When URLs Contain Different Content

I think that we have great content on the lesson pages which should be used as we discussed with Andrew.
Maybe the callout for when a user lands on a lesson directly "Start from the beginning!" or similar would be enough to help a user and even though it's not ideal, we'd still get more inbound traffic because of the great content lessons provide.

I suggest:

 - Adding lessons pages to the sitemap
 - Adding callout to start the tutorial from the beginning

Would like Andrew's input again on this.

terichadbourne:
[...]

PR #521, opened 2020-08-07
Fix: more seo updates

zebateira:
This PR tackles the SEO issues described in #506.

Also made two other changes:

- improve page loading performance for some pages: create javascript chunks for each page type using `import` statement and `webpackChunkName` directive
- remove old tutorials [redirect code](https://github.com/ProtoSchool/protoschool.github.io/pull/521/files#diff-8653d8e05b6a35b56511fda12b92ed06L237)
  - because we now have server side routing, we need to add the redirects to the `netlify.toml` file. I've created another issue to tackle this: #520 - this affects our custom redirects and tutorial url redirects.


src/routes.js
```diff
@@ -12,30 +16,39 @@ const types = {
...
-  { type: types.STATIC, path: '/contribute/' },
-  { type: types.STATIC, path: '/news/' },
-  { type: types.STATIC, path: '/tutorials/' },
+  { type: types.STATIC, loc: '/', priority: 1, changefreq: 'weekly', lastmod },
+  { type: types.STATIC, loc: '/tutorials/', priority: 0.9, changefreq: 'weekly', lastmod },
+  { type: types.STATIC, loc: '/events/', priority: 0.8, changefreq: 'weekly', lastmod },
+  { type: types.STATIC, loc: '/news/', priority: 0.7, changefreq: 'monthly', lastmod },
+  { type: types.STATIC, loc: '/host/', priority: 0.6, changefreq: 'monthly', lastmod },
+  { type: types.STATIC, loc: '/build/', priority: 0.6, changefreq: 'monthly', lastmod },
+  { type: types.STATIC, loc: '/contribute/', priority: 0.6, changefreq: 'monthly', lastmod },
```
zebateira, line 49: @andrew what do you think about these values for the sitemap?
We were wondering if `/tutorials` is more "important" than `/` for example.

andrew: LGTM, for a site with this number of pages it won't really have much of an impact no matter what you set it to

olizilla:
> described in #506 + others:

To help me review, please list out all the issues that this PR fixes.

zebateira:
> To help me review, please list out all the issues that this PR fixes.

Sorry for the confusion. I added the remaining issues I fixed to #506 and updated the description of this PR as well.

package.json
```diff
@@ -17,7 +17,7 @@
     "scripts:build:data": "node scripts/commands/build-data",
     "scripts:googleapis-generate-token": "node scripts/commands/googleapis-generate-token",
     "scripts:wizard": "node scripts/commands/wizard/index.js",
-    "sitemap": "rm public/sitemap.xml && vue-cli-service sitemap",
+    "sitemap": "rm public/sitemap.xml ; vue-cli-service sitemap",
```
olizilla, line 20: nitpick, but if you want `rm` to succeed even if the file is missing, then use `rm -f`. If `rm` fails for some other reason, then you probably dont want to run `vue-cli-service sitemap`

```suggestion
    "sitemap": "rm -f public/sitemap.xml && vue-cli-service sitemap",
```

public/sitemap.xml
```diff
@@ -1 +1 @@
...
...</changefreq><priority>0.6</priority></url><url><loc>https://proto.school/contribute/</loc><lastmod>2020-08-06</lastmod><changefreq>monthly</changefreq><priority>0.6</priority></url><url><loc>https://proto.school/data-structures/</loc><lastmod>2020-08-06</lastmod><changefreq>monthly</changefreq><priority>1.0</priority></url><url><loc>https://proto.school/basics/</loc><lastmod>2020-08-06</lastmod><changefreq>monthly</changefreq><priority>1.0</priority></url><url><loc>https://proto.school/blog/</loc><lastmod>2020-08-06</lastmod><changefreq>monthly</changefreq><priority>1.0</priority></url><url><loc>https://proto.school/mutable-file-system/</loc><lastmod>2020-08-06</lastmod><changefreq>monthly</changefreq><priority>1.0</priority></url><url><loc>https://proto.school/regular-files-api/</loc><lastmod>2020-08-06</lastmod><changefreq>monthly</changefreq><priority>1.0</priority></url><url><loc>https://proto.school/anatomy-of-a-cid/</loc><lastmod>2020-08-06</lastmod><changefreq>monthly</changefreq><priority>1.0</priority></url><url><loc>https://proto.school/verifying-storage-on-filecoin/</loc><lastmod>2020-08-06</lastmod><changefreq>monthly</changefreq><priority>1.0</priority></url></urlset>
```
olizilla, line 1: Any reason not to exclude this file from git and have it autogenerated during the build? It's normally a footgun to to have generated resources commited to the repo.

src/components/Navigation.vue
```diff
@@ -31,8 +31,8 @@
         <!-- standard nav  -->
         <div v-else class="flex-auto link fw6 f5 db bb border-aqua">{{currentPage}}</div>
         <button @click="toggleHamburger" class="menu-toggle button-reset bg-transparent b--transparent">
-          <img v-if="isHamburgerClosed" src="../static/images/burger.svg"/>
-          <img v-else src="../static/images/close.svg"/>
+          <img v-if="isHamburgerClosed" src="../static/images/burger.svg" alt="open navigation menu" />
+          <img v-else src="../static/images/close.svg" alt="close nativation menu" />
```
olizilla, line 35:
```suggestion
          <img v-else src="../static/images/close.svg" alt="close navigation menu" />
```

src/pages/Lesson.vue
```diff
@@ -25,24 +25,23 @@
...
         :createTestTree="logic.options.createTestTree"
         :lessonId="parseInt(lessonId, 10)"
-        :lessonTitle="lesson.title"
+        :lessonTitle="!!lesson && lesson.meta.title"
     />
     <MultipleChoiceLesson
-        v-else-if="lesson.type === 'multiple-choice'"
-        :text="text"
+        v-else-if="!!lesson && lesson.meta.type === 'multiple-choice'"
+        :text="!!esson && lesson.html"
```
olizilla, line 32: I dont know what this does but you probably dont want `esson` here. Can you add a cypress test to show the changes you expect to see from these?

src/utils/tutorials.js
```diff
@@ -234,28 +234,4 @@ export function isLessonPassed (tutorial, lesson) {
   return !!localStorage[`passed/${tutorial.url}/${lesson.formattedId}`]
 }
```
olizilla, line 236: If these redirects are needed, do you plan to land the netlify config to recreate them before merging this PR?

public/sitemap.xml
zebateira: Ah sounds good 👍 
Was thinking it would be nice to have a local committed version, but we can always run the build command to test the final result.

src/pages/Lesson.vue
zebateira: The component would crash always on the `lesson.meta.title` first, so I only added this in case vue internals change.

```suggestion
        :text="!!lesson && lesson.html"
```

src/utils/tutorials.js
zebateira: This feature is not being used still, but it may be in the future.
Meaning, only when we have `redirectUrls` in any of the (tutorials](https://github.com/ProtoSchool/protoschool.github.io/blob/main/src/static/tutorials.json#L4) will this feature be used.

But regarding the issue #520, I've changed its goal to actually include client side redirects to be used as fallback while the server configuration is not set yet. This seems like a good approach to not rely on the server config for the redirects to work, but instead, rely on them only for the SEO.
I'm working on a PR for this already.

So the removal of this code here will not break anything.

src/pages/Lesson.vue
olizilla: Would it be possible to write a cypress test for the logic in this file? it's not immediately obvious how it all fits together and it wasn't clear that one of the conditions was broken. Would be great to have a test that fails rather than relying on human eyes to catch typos.

zebateira: Ah yes, will do 👍

zebateira: Pushed a test for this here: https://github.com/ProtoSchool/protoschool.github.io/pull/521/commits/eaa1c236ad58795be7529931047344e6154b13a2

Good catch on this. Lesson's with multiple choice were not rendering the lesson content and we weren't testing for that.

At first I thought I could simply do snapshot testing, but then every time a typo or any other content parts change in the future, the tests would fail.
Another approach I tried was to assert the rendered html, but because of the sanitized result of the html of vue's `v-html`, it was hard to reproduce the same resulting html.

For this test assertion I ended up adding the resulting html from the markdown (result of `marked()`) to a `data-cy-text` attribute so that we could assert the text ended up correctly in the element (`<div class="lesson-text" v-html="text" :data-cy-text="text">`). This way we can check the loaded markdown string is the same as the one rendered in `v-html`.

Let me know if you know of a better way to test this.

public/sitemap.xml
zebateira: Done!

src/utils/tutorials.js
zebateira: PR to add pre-rendered pages for redirects added here: https://github.com/ProtoSchool/protoschool.github.io/pull/525

PR #525, opened 2020-08-12
Fix: prerender redirect pages 

### LINKED ISSUE #520: Fix: prerender redirect pages so redirects can be handled on client side

zebateira:
Closes #520 
Blocked by #521 since we needed the latest SEO changes.

To test this PR:

```
npm run build
npm run cy:serve
```

Check `localhost:3000/chapters` redirects to `/events`. Before this PR it would return 404.


terichadbourne:
> Check localhost:3000/chapters redirects to /events. Before this PR it would return 404.

I've tested this **locally** as described and it redirects from both localhost:3000/chapters and localhost:3000/chapters/ (with and without trailing slash) to localhost:3000/events (**no trailing slash**). 

However, I also tested on the **live ProtoSchool site**, which now has #521 merged (but not this PR), and there it already  redirects localhost:3000/chapters and localhost:3000/chapters/ (with and without trailing slash) to localhost:3000/events/ (**with trailing slash**).

So I'm wondering if the issue was accidentally fixed in a different way previously or if there's a better way for me to test what's changed here. 

Was going to wait for the Netlify preview and test that as well but the failed Netlify build shown here looks successful in the detailed contents of the [deploy log for the preview](https://app.netlify.com/sites/protoschool-prod/deploys/5f3d4bcfc7c32800b4400213) end in "Netlify Build Complete." Sigh. Perhaps @andyschwab can help me figure out why Netlify is complaining about this one?

terichadbourne:
Just found a note I made to myself that this PR is meant to provide a client-side fallback redirect when we switch over to deploying on Fleek, before they've set up the redirects there that we've already set up on Netlify. So I'm no longer concerned that it doesn't improve the user experience outside of Fleek. 

I just tested it on this version of the Fleek preview versus one on another branch. On this one it worked as intended and on the other, rather than deliver a 404 page, it did this, which is way worse: 

`ipfs resolve -r /ipfs/QmVviawHGkXBtNG55Tw2AFgHt5uzUHsYGfozAAzyf3jjYX/chapters/: no link named "chapters" under QmVviawHGkXBtNG55Tw2AFgHt5uzUHsYGfozAAzyf3jjYX`

Therefore I declare this PR A++ and I will merge it if by some miracle Netlify ever actually lets me. 🤦‍♀️

PR #534, opened 2020-08-18
fix: do not use trailing slashes in local storage item ids

### LINKED ISSUE #530: Caching issues with trailing slashes in URLs
Unfortunately there is a bug in the first page of lesson 1. If the incorrect option is selected once even selecting the correct option will not allow progression to the next Q.  This happens even if the browser is closed and link reopened. See attached photo:
AB0D28E6-5960-4C8C-888E-C312286A35C0

zebateira:
Fixes #530

Any new progress with the trailing slash will not be taken into account, but previous progress without the slash will still be taken into account. I don't think this will be a problem since advancing any lesson was working, so effectively, no progress will be lost, but I thought I'd still mention it.

### DIFF
--- a/cypress/integration/tutorials.spec.js
+++ b/cypress/integration/tutorials.spec.js
@@ -486,3 +486,47 @@ function advanceThroughLessons (tutorialId) {
     }) // end this lesson
   }) // end loop through standard lessons, landing on resources page
 } // end advanceThroughLessons
+
+describe('SHOULD ADVANCE LESSONS WITH TRAILING SLASHES', () => {
+  it('should advance text lessons', () => {
+    cy.visit('/data-structures/01/')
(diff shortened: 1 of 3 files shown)


terichadbourne:
I'm getting the expected behavior on both trailing slash and slash-free versions for each lesson type now. Existing tests are still passing, and I know you're working on adding new ones for the trailing slash versions. Thank you @zebateira!

> Any new progress with the trailing slash will not be taken into account, but previous progress without the slash will still be taken into account. I don't think this will be a problem since advancing any lesson was working, so effectively, no progress will be lost, but I thought I'd still mention it.

Yeah, I think the most valuable thing someone would lose would be cached code they'd written but hadn't been able to see was passing, but I think you're right that it would've only happened in a single lesson.

PR #547, opened 2020-09-04
fix: document metadata was not updating between some pages

### LINKED ISSUE #526: Bug: incorrect metadata rendering
[This tweet](https://twitter.com/filecoin/status/1292919017833889792?s=21) links to the filtered Filecoin course but the Twitter card shows as the homepage (and not even with the correct homepage title):

8C00D7CC-F366-44F0-9F7D-86F7DBFBD9E9

@zebateira Any idea what could be going wrong?

zebateira:
`head` was not being called on every page change.

Change on some pages from

```js
head: head()
```

to 
```js
head() {
    return head()
}

Closes #526

### DIFF
--- a/src/pages/Build.vue
+++ b/src/pages/Build.vue
@@ -83,6 +83,8 @@ export default {
   components: {
     Header
   },
-  head: head()
+  head () {
+    return head()
+  }
 }
 </script>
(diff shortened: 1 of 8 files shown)


terichadbourne:
Awesome! Working in both Netlify and Fleek previews.

PR #551, opened 2020-09-10
fix: remove trailing slash on vue router to make countly data consistent

zebateira:
I noticed that in countly we were getting a lot of data points with inconsistent urls that include trailing slashes and others that don't.
Examples:

 - Events: https://countly.proto.school/dashboard#/5d25fd49db14b4435e4432e2/analytics/events/key/tutorialFeedbackSurveyCompleted
- Page Views (notice the `/data-structures` url has a duplicated entry with `/data-structures/`

image

This PR will remove the trailing slash on client-slide. Netlify/Fleek trailing slash redirect will not be affected and will still happen.

### How it works

By using `vue-router`'s `beforeEach` function, we can remove the trailing slash from window location if it's present, and only the client side will be affected.
So the redirect from Netlify/Fleek to add the trailing slash will still happen, but then `vue-router` kicks in and will remove it, so that our application state does not contain the trailing slash. Meaning Countly and other vue router accesses will be correct: no trailing slash.

This fixes a lot of Countly events being logged incorrectly and also prevents future bugs that depend on the route.

### DIFF
--- a/cypress/integration/redirects.spec.js
+++ b/cypress/integration/redirects.spec.js
@@ -16,7 +16,7 @@ describe('REDIRECTS', () => {
     assertHashRouteRedirect('/data-structures/resources')
     assertHashRouteRedirect('/data-structures/01')
     assertHashRouteRedirect('/basics')
-    assertHashRouteRedirect('/basics/')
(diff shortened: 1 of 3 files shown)


terichadbourne:
Looks great!

PR #556, opened 2020-09-14
fix: force page refresh on ChunkLoadError

### LINKED ISSUE #555: Bug: ChunkLoadError Loading chunk failed
**Describe the bug**
When a user comes back to a tab with ProtoSchool open after a deployment, they might get this `ChunkLoadError` error because the chunk's name does not exist anymore (because of cache busting hashes in the filenames).

**Screenshots**
image

**To reproduce**
Make a prod build, open tab, change content in the events page (e.g.), build again and click the events link in the navigation in the previous opened tab. It should replicate the issue.

**Additional context**
https://dev.to/ianwalter/fixing-chunkloaderror-3791

Thank you @andyschwab and @autonome for looking into this 🙏 

**Proposed Solution**
Possible solution for this would be to add an error handler to catch the `ChunkLoadError` error and if it happens, simply reload the page.

zebateira:
When a user comes back to a tab with ProtoSchool open after a deployment, they might get this ChunkLoadError error because the chunk's name does not exist anymore (because of cache busting hashes in the filenames).

The fix for this is to force the browser to reload.
In addition to that, we are navigating the user to the route they wanted to change to, so that instead of simply reloading the page, the user actually navigates to the correct page.

Closes #555 

### Refs:

 - https://dev.to/ianwalter/fixing-chunkloaderror-3791

### Steps to test this PR:

1. run prod build and server it: `npm run build && npm run cy:serve`
2. open localhost:3000 without cache enabled (dev tools > network > disable cache: on)
3. change content in the Events page. Open `src/pages/Events.vue` file, and change the content (add `test` anywhere in the template)
4. close and run prod build again: `npm run build && npm run cy:serve`
5. without refreshing, click on the events link in the navigation
6. the website should reload and land on the `/events` page

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -47,12 +47,31 @@ router.afterEach((to) => {
   window.Countly.q.push(['track_pageview', to.path])
 })
 
+// To be used in the onError handler
+let nextRoute = null
+
 router.beforeEach((to, from, next) => {
+  nextRoute = { ...to }
+
   // Remove trailing slash on client side only
   if (to.path !== '/' && to.path.endsWith('/')) {
     next({ path: to.path.substring(0, to.path.length - 1), replace: true, query: to.query })
   } else {
     next()
+  }
+})
+
+router.onError(error => {
+  // https://github.com/ProtoSchool/protoschool.github.io/issues/555
+  // When a user comes back to a tab with ProtoSchool open after a deployment,
+  // they might get this ChunkLoadError error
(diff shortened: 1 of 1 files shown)


terichadbourne (approved):
Thanks for the step-by-step testing instructions. I was able to replicate the error by following them on the main branch and get rid of that error by taking the steps described on this new branch. Looks good!

PR #560, opened 2020-10-01
Create multiple choice quizzes using ProtoWizard CLI

### LINKED ISSUE #445: Feature: Build multiple choice questions in ProtoWizard
After we get our CLI published (#440), let's circle back and make it possible to build a multiple-choice quiz through the CLI instead of just by editing the JS file.

terichadbourne:
Closes #445. 

Adds the ability to create a quiz (edit the JS file for a multiple-choice lesson) from the ProtoWizard for those less comfortable with digging into the code. 

Need to discuss building tests. 

Also need to update docs to reflect this new option.


scripts/commands/wizard/lesson.js
```diff
@@ -27,6 +24,7 @@ function logLessons (message, lessons) {
 async function createLessonIntro ({ createResource, createTutorial, createLesson }) {
   // determine new tutorial number
   log.info("Let's add lessons to your tutorial.")
+  // TODO: Add createQuiz below? But it says it's not defined when I try.
```
zebateira, line 27: this is done?

scripts/commands/wizard/quiz.js
```diff
@@ -0,0 +1,187 @@
...
+// TODO confirm overwrite existing quiz if there's real data
+// DONE prompt for question
+// DONE prompt for correct answer (text, success message)
+// DONE prompt repetitively for 2-4 wrong answers (text, failure message)
+// DONE randomly sort order of wrong answers
+// DONE store data to js file (remember to deal with quotes correctly)
+// DONE log randomized questions?
+// DONE show link to preview
+// DONE show which file to go to to edit text
+// DONE create another quiz OR exit
```
zebateira, line 18: all done, so we can remove these?

scripts/commands/wizard/quiz.js
```diff
@@ -0,0 +1,187 @@
...
+  promptRepeat,
+  logPreview,
+  logCreateLater
+} = require('./utils.js')
+
+// *** QUIZ CREATION ***
+
+async function createQuiz (tutorial, lesson, { createLesson, createTutorial, createResource }) {
+  log.info("Now it's time to write the question for your multiple-choice quiz and provide answer choices, with positive or negative feedback for each. Wrong answers, and the feedback associated with them, are a great way to address common misconceptions about the topic. Be sure to make your feedback as helpful as possible, to guide the learner to the right choice. You'll need to create 1 correct answer and 2-3 incorrect answers. (I'll take care of randomizing their order later.")
+  const responses1 = await inquirer.prompt([
```
zebateira, line 39: we can rename this to `responses` and on the bellow variable `responses2` we can simply reuse the initial one.

scripts/commands/wizard/quiz.js
```diff
@@ -0,0 +1,187 @@
...
+      if (shouldOverridePristineQuiz || shouldOverrideQuiz) {
+        return createQuiz(tutorial, lesson, { createLesson, createTutorial, createResource })
+      } else {
+        logEditQuizManually(tutorial, lesson)
+      }
+    }
+  } else {
+    log.info('No worries. Please run the ProtoWizard again to create your tutorial and lesson. Once you have the necessary files, you can run this script again to build your quiz.')
+  }
+} // end createQuizIntro
```
zebateira, line 145: Let's add some spacing here to make it more readable? I suggest the following:

```suggestion
async function createQuizIntro ({ createLesson, createTutorial, createResource }) {
  log.info("Let's create a multiple-choice quiz! This will only work if you've already created the tutorial and lesson files.")

  if (await promptFilesReady()) {
    const tutorial = await selectTutorial('quiz', { createTutorial, createResource, createLesson, createQuiz })
    const lesson = await selectLesson(tutorial)
 
    if (lesson) { // skip if value is null because there weren't multiple choice lessons in tutorial
      logList(`Great! You've chosen to build a quiz for`, [`Tutorial: ${tutorial.title}`, `Lesson: ${lesson.title}`])

      const shouldOverridePristineQuiz = api.lessons.isQuizPristine(tutorial, lesson) && await promptOverwritePristineQuiz(tutorial, lesson)
      const shouldOverrideQuiz = !shouldOverridePristineQuiz && await promptOverwriteQuiz(tutorial, lesson)

      if (shouldOverridePristineQuiz || shouldOverrideQuiz) {
        return createQuiz(tutorial, lesson, { createLesson, createTutorial, createResource })
      } else {
        logEditQuizManually(tutorial, lesson)
      }
    }
  } else {
    log.info('No worries. Please run the ProtoWizard again to create your tutorial and lesson. Once you have the necessary files, you can run this script again to build your quiz.')
  }
} // end createQuizIntro
```

zebateira: I would also do this with the rest of the file. Just adding some spaces after variable declaration blocks helps, as well as after logs.

scripts/commands/wizard/resource.js
```diff
@@ -37,6 +37,7 @@ async function createResourceIntro ({ createLesson, createTutorial }) {
   // determine new tutorial number
   log.info("Let's add resources to your tutorial.")
 
+  // TODO: Add createQuiz below? But it says it's not defined when I try.
```
zebateira, line 40: remove this?

src/api/modules/lessons.js
```diff
@@ -83,6 +83,35 @@ function get (tutorial, lessonId) {
  * @example
  * api.lessons.create(tutorial, { title: 'Lesson title', type: 'text' })
  */
+
+function updateQuiz (tutorial, lesson, data) {
```
zebateira, line 87: We need to move this to before the doc comment block of the create function.
we need to add more docs comments to each in the future, but for now lets just make sure the create function comments are right before the create function.

zebateira (changes requested):
Looking good 👍

terichadbourne:
Thanks for the feedback, @zebateira!  I believe I've gotten everything you mentioned, and I've also updated the Developing Tutorials guide with new instructions and an updated screenshot. Let me know what you think. 

Looks to me like the build errors are a result of us not having updated the tests yet, which we'll address tomorrow.

terichadbourne:
@zebateira Since we last met I've been able to fix the Jest test for 4.3 but am still have trouble with 4.1 and 4.2. In both cases it seems like the ProtoWizard itself is working fine and it's the tests themselves not doing what we intend. 

In 4.1, I fixed a few formatting issues already and believe I'm down to spacing/tabbing differences. The formatting between the substring and the string is identical to the eye and not to the computer. I've tried a variety of spacing, copying code over to the same editor to check number of spaces, etc. but haven't yet been able to make these match each other. Any suggestions? 
image

image

_____________________

For 4.2, our test is somehow comparing the expected final data to the halfway-through-the-process meant-to-be-overwitten data instead of the final product. 
image


I tried going through the ProtoWizard again and confirmed that all the steps matched the runners.ProtoWizard in the 4.2 test: 
image

(First I run the ProtoWizard to create a new tutorial with one new multiple choice lesson and exit fully from the ProtoWizard before re-entering the ProtoWizard and taking the steps shown there.) 

I watched the JavaScript file to confirm it was indeed updating during the process. During my second run through updating the quiz, the file is still on the first set of submitted choices as it should be: 
image

After I tell it I'm done adding choices the 2nd time around, the file is updated as it should be: 
image

I do get what I expected to in the terminal, as I create one set of quiz data that should be overwritten and then another set that I overwrite it with: 
image
image

The snapshots, however, seem to bail out early... it seems to only collect snapshots during the first run through creating a quiz, even though the user doesn't exit the ProtoWizard in the middle. (This was with me having deleted the snapshot file and then run `npm run test jest` a single time.) It only gets as far as where you select the lesson to build your quiz for the first time through and then appears to stop collecting snapshots. 

image

Can you think of a reason it would bail in the middle of the runner? My only guess is that I've set up a race condition and it's trying to run lines 113-118 at the same time as lines 81-111. If that's true, it's probably happening elsewhere in this file as well and just showing up here because it's the longest runner. I checked the tests you created to confirm they're written the same way as mine, and they seem to be, but those runners are shorter than all of mine. Thoughts?

terichadbourne:
This is now ready for review, @zebateira. I've fixed the remaining issues that were causing Jest tests to fail (thanks for the help debugging!), removed unused console logs, and updated some wording to simplify some logging and make the ProtoWizard more personable. 

Would you mind adding to the placeholders I started in `scripts/README.md` to add a little more info for future contributors who need to play with Cypress and Jest?

jest/helpers/asserts.js
```diff
@@ -1,5 +1,26 @@
...
+
+// returns true if log includes a string from the excludedPhrases array
+function isExcluded (log) {
+  for (let phrase of excludedPhrases) {
+    if (log.includes(phrase)) {
+      return true
+    }
+  }
+  return false
+}
```
zebateira, line 16: Just a suggestion to make it simpler:

```suggestion
function isExcluded (log) {
  return excludedPhrases.some(phrase => log.includes(phrase))
}
```

jest/helpers/asserts.js
```diff
@@ -1,5 +1,26 @@
...
+
+// add log to snapshot UNLESS it contains an excluded phrase
 function assertLogSnapshot (log) {
-  expect(log).toMatchSnapshot()
+  if (!isExcluded(log)) {
+    expect(log).toMatchSnapshot() // expect if log isn't excluded from exertion
```
zebateira, line 21: I don't understand this comment because of the word exertion. I think it's just a language issue that I don't understand, but just wanted to make sure that this is the case, because I think you meant `assertion` instead of `exertion`?

scripts/__tests__/protowizard/quizzes.spec.js
```diff
@@ -0,0 +1,163 @@
...
+        { latestTutorial: true }, // add to latest (test) tutorial
+        { lesson: tutorial.lessons[0] }, // select first (test) lesson
+        { confirm: true }, // overwrite pristine quiz
+        { question: 'Question to overwrite', // question
+          correctAnswer: 'Correct answer to overwrite', // correct answer
+          correctFeedback: 'Correct feedback to overwrite' }, // correct feedback
+        { incorrectAnswer: 'Incorrect answer to overwrite', // incorrect answer
+          incorrectFeedback: 'Incorrect feedback to overwrite' }, // incorrect feedback
+        { confirm: false }, // don't add another answer
+        // ***THE STATE OF THE JS FILE RIGHT HERE (WITH DATA MEANT TO BE OVERWRITTEN) IS WHAT JEST IS COMPARING TO WHEN IT FAILS
```
zebateira, line 94:
```suggestion
```

scripts/commands/wizard/quiz.js
```diff
@@ -0,0 +1,175 @@
...
+    log.info(`Answer (Incorrect): "${wrongAnswer.answer}"`)
+    log.info(`Feedback: "${wrongAnswer.feedback}"`)
+
+    choices.push(wrongAnswer)
+
+    log.info(`You currently have 1 correct answer and ${choices.length - 1} wrong answer${choices.length > 2 ? 's' : ''}. (I recommend providing 2-3 incorrect options.)`)
+
+    if (!(await promptRepeat('wrong answer'))) {
+      askAgain = false
+    }
```
zebateira, line 90: Just a suggestion to simplify the logic:

```suggestion
    askAgain = await promptRepeat('wrong answer') // prompt to repeat adding a wrong answer
```

scripts/commands/wizard/quiz.js
```diff
@@ -0,0 +1,175 @@
...
+  log.info("Let's create a multiple-choice quiz! This will only work if you've already created the tutorial and lesson files.")
+
+  if (await promptFilesReady()) {
+    const tutorial = await selectTutorial('quiz', { createTutorial, createResource, createLesson, createQuiz })
+    const lesson = await selectLesson(tutorial)
+
+    if (lesson) { // skip if value is null because there weren't multiple choice lessons in tutorial
+      logList(`Great! You've chosen to build a quiz for`, [`Tutorial: ${tutorial.title}`, `Lesson: ${lesson.title}`])
+
+      const overwrite = api.lessons.isQuizPristine(tutorial, lesson)
```
zebateira, line 118: Just a nick-pick. Using `should*` indicates that it is boolean and that we will perform the override or not depending on its value. 

```suggestion
      const shouldOverride = api.lessons.isQuizPristine(tutorial, lesson)
```

scripts/commands/wizard/quiz.js
```diff
@@ -0,0 +1,175 @@
...
+        ? await promptOverwritePristineQuiz(tutorial, lesson)
+        : await promptOverwriteQuiz(tutorial, lesson)
+
+      if (overwrite) {
```
zebateira, line 122:
```suggestion
      if (shouldOverride) {
```

scripts/commands/wizard/utils.js
```diff
@@ -57,22 +61,44 @@ async function selectTutorial (newItemType, { createTutorial, createResource, cr
...
     } else {
-      tutorial = await createTutorial({ createLesson, createResource }, { skipPromptLesson: true })
+      tutorial = await createTutorial({ createLesson, createResource, createQuiz }, { skipPromptLesson: true })
     }
   }
 
   return tutorial
 }
 
+async function selectLesson (tutorial) {
```
zebateira, line 83: Maybe we should rename it to `selectMultipleChoiceLesson`?
The `selectTutorial` function only selects one tutorial and does not discriminate between the tutorial types.

src/api/modules/lessons.js
```diff
@@ -52,6 +52,9 @@ function get (tutorial, lessonId) {
       if (lesson.type !== 'multiple-choice') {
         lesson.files.challengeMarkdown = files.getChallengeMarkdownPath(tutorial, lessonId)
       }
+      // TODO else if mult choice, lessons.choices = choices
+      // require JS file to get choices object  (path at line 50)
+      // but we can't actually require it so...  ¯\_(ツ)_/¯
```
zebateira, line 57:
```suggestion
```

zebateira (changes requested):
Good job 👏
Just a few nick-picks and TODO cleaning 🧹

zebateira (approved):
🎉

Issue #386, opened 2020-02-18
Upgrade Cypress to 4.0 to support cross-browser testing


terichadbourne:
Cypress recently announced v 4.0 which supports testing in MS Edge and Firefox in addition to Chrome. However, it has breaking changes so we'll need to explore thoroughly before upgrading. 

- [Announcement](https://cypress.io/blog/2020/02/06/introducing-firefox-and-edge-support-in-cypress-4-0/)
- [Migration Guide](https://docs.cypress.io/guides/references/migration-guide.html#Migrating-to-Cypress-4-0)
- [Cross-Browser Testing Guide](https://docs.cypress.io/guides/guides/cross-browser-testing.html#Continuous-Integration-Strategies)

zebateira:
We've upgraded to `v5` on #562, so next I'll add more browser testing

PR #562, opened 2020-10-19
Chore: Upgrade Cypress to version 5

zebateira:
This PR upgrades Cypress to version 5.

I'll make a separate PR to add testing on other browsers on CI (Fleek) because we might need to update the docker image we are using to include the other browsers. So this only addresses #386 partially.

Refs:

- [Cypress Migration Guide](https://docs.cypress.io/guides/references/migration-guide.html#Mocha-upgrade)

### DIFF
--- a/cypress/integration/tutorials.spec.js
+++ b/cypress/integration/tutorials.spec.js
@@ -6,6 +6,11 @@ import courses from '../../src/static/courses.json'
 
 const tutorials = _.omitBy(tutorialsList, tutorial => tutorial.hidden)
 
+const fixtures = {
+  png: 'favicon.png',
+  json: 'example.json'
+}
+
 // lessons to use as samples when testing functionality only once per lesson type
 const testLessons = {
   code: {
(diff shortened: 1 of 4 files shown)


zebateira:
Fleek is timing out because the build is exceeding 15mins. Each code submission is taking longer than usual. This might be related the environment IPFS is running in.
I _think_ it was a bit slow before, but now with cypress@5 seems to be even slower.

zebateira:
> This might be related the environment IPFS is running in.

This does not seem to be the case. I'm running it locally and it's working perfectly.

image

The first tests take longer on fleek.

### Local

image

### Fleek

image


Roughly, double the time.

This only started to happen when I upgraded to to cypress@5, but it only happens in Fleek. I'll try to replicate this on the docker image.

terichadbourne:
Thanks for exploring this with the Fleek team @zebateira. Would it be more helpful for me to proof the PR now or wait until you've resolved that conversation?

zebateira:
> Thanks for exploring this with the Fleek team @zebateira. Would it be more helpful for me to proof the PR now or wait until you've resolved that conversation?

Proofing this PR shouldn't be necessary - if it passes on fleek, should be ok. So that's exactly what I need to fix ahah.

zebateira:
By using another docker image on fleek (`fleek/puppeteer:cypress`) the build went back to normal values - merging this as is so we can start working on #386

PR #565, opened 2020-10-29
chore: add firefox testing to ci build command

### LINKED ISSUE #386: Upgrade Cypress to 4.0 to support cross-browser testing

zebateira:
This PR adds more browsers to the Cypress test suite.

Main changes:

- Upgrade `cypress` from `5.4.0` to `5.5.0`
- Upgrade `cypress-file-upload` from `3.5.3` to `4.1.1`
- Change `uploadFiles` property to only contain an array (boolean of value `false` is now `[]`).
- Fix "Reset files" link button on tests (only started to fail with the `cypress-file-upload`'s new version
    - we were missing resetting the input value
- Update cypress timeout configs: remove `responseTimeout` and reduce `defaultCommandTimeout` to `30s`.
    - No need to have such large timeouts now - they should pass just fine with `30s` due to the new Cypress version which introduces some performance fixes.

Closes #386


zebateira:
@terichadbourne should be ready for review now :)
To test this one can run the command: `npm run build && npm run cy:run:ci`
It should run all the tests both on chrome and firefox (~20mins)

terichadbourne (approved):
Looks good! As already discussed, I did have to extend the timeout period to get this to pass, but we'll merge as is and then fix the timeout if we have continued problems.

PR #566, opened 2020-11-05
Implement page-based social tiles

terichadbourne:
This PR uses Vue-Meta to provide distinct social tiles for each page that can have its own metadata. 

TODO: 
- [x] Fix bug where pages end up with two `twitter:image` fields, the correct one and the generic one we thought it would overwrite. 
- [x] Replace all images except for the 3 home images b/c they have a 1px wide white line on the right side: 
   - [x] 7 tutorial cards
   - [x] build, contribute, host, events a/b/c
- [x] Add a news card


terichadbourne:
I've made some final updates to image files as planned. @zebateira could you please give a final proof and merge if everything looks good?

zebateira (approved):
great job 🎉

PR #571, opened 2020-11-16
Surface survey more prominently

terichadbourne:
In order to make the survey more prominent and increase completion rates, we've been asked to move it above the resources. If this happens with Option B, the resources can't be seen at all on page load:

image

With Option A, you can see that the resources exist below the survey: 

image

Unfortunately the smaller version that looks better here, Option A, has a 6.3% completion rate, while Option B has an 8.4% completion rate. 

@zebateira can you please make the following edits? 

- [x] Dismiss congrats message automatically if you click to share on twitter
- [x] Add a manual dismiss option for the congrats message
- [x] Adjust the progress bar on Option A to make it more obvious there are only 3 questions (increasing completion)
- [x] Adjust Option B to make questions disappear as you answer them so you can see the resources appear sooner

We'll then continue testing improved option A and improved option B, both at the top.


zebateira:
@terichadbourne done with the changes (see previews bellow). Let me know what you think.

# Dismiss Tutorial Message Callout on twitter share and add dismiss button


tutorial callout 1
tutorial callout dismiss


 # Option A

option A


# Option B

option B


# Countly Events

Screenshot 2020-11-20 at 11 07 11 AM

terichadbourne:
Thanks for your work on this @zebateira! 


Option B is working for me if I answer the questions in order but acting up if I don't.  If I answer the questions out of order, sometimes all unanswered ones disappear, sometimes two disappear at once, including an unanswered one. And I end up with just the header  instead of the CTA for the profile survey. image
Some of the sequences that have caused this: 

- answer number 3,  1 + 3 go away, answer2, 2 goes away and header stays 
- answer number 2,  1 + 2 go away,  answer 3,  3 goes away and  header stays
- answer number 1,  1 goes away,  answer 3,   2 + 3 go away and header stays

My ideal would be that completing all 3, rather than completing the 3rd one, would make the whole box disappear. And if you completed just one, no matter which it was, that one would disappear and the rest remain. Once all 3 were completed, the contents would be replaced by the CTA. 

----

I love the intent of the pills for progress in Option A, but before answering the first question, I'm not familiar enough with this UI to know I'm looking at a progress indicator and it could look like an underlining mistake on the header. The only think I can think of to counteract that in a similar format is making the pills pointy, like arrows to the right pointing into each other, but I don't think the look would fit here. And of course something more literal like a Question 1 of 3  or 1 or 3 in a corner would be an option, but less pretty. Maybe worth taking to the design group for ideas?
image

----

I was expecting us to send a profileSurveyClick event to Countly when I clicked to go to the Google Form, but it's not registering the test version of ProtoSchool. We'll want this stat to divide by the number of tutorial feedback survey completions to get a percentage of folks who give us tutorial feedback who continue on to give us the profile survey feedback. Although as I think of it this makes more sense as a percentage of itself, since you'd only fill out the survey profile once per person. Has this already been accounted for and I'm looking at the wrong event name? I don't actually see the stat at all in the real dashboard, so maybe we pulled it thinking we'd just take the number from the form submissions? Do you recall, @zebateira? 

---- 

Likewise, I was expecting a sucessLinkClick when clicking to draft a Tweet, but it doesn't produce an event in the test dashboard on Countly and the event doesn't exist in the real dashboard. Do you recall what happened here?

zebateira:
Updated with fixes for option A & B.

## Option A

Screen Cast 2020-11-24 at 4 52 32 PM

## Option B

Screen Cast 2020-11-24 at 4 53 29 PM

zebateira:
> I was expecting us to send a profileSurveyClick event to Countly when I clicked to go to the Google Form, but it's not registering the test version of ProtoSchool. We'll want this stat to divide by the number of tutorial feedback survey completions to get a percentage of folks who give us tutorial feedback who continue on to give us the profile survey feedback. Although as I think of it this makes more sense as a percentage of itself, since you'd only fill out the survey profile once per person. Has this already been accounted for and I'm looking at the wrong event name? I don't actually see the stat at all in the real dashboard, so maybe we pulled it thinking we'd just take the number from the form submissions? Do you recall, @zebateira?

I debugged this and fixed it. Basically we were not sending the event correctly, so we have only logged it in our testing app (in countly) and so far we have zero logs on prod as you mentioned.
The problem was that we added the `PROFILE_SURVEY_CLICK` event to the old static countly file which is not being used anymore. So I removed it and moved the constant to the correct `utils/countly.js` file.
It should be logging the event now on our testing app.

zebateira:
@terichadbourne 

> Likewise, I was expecting a sucessLinkClick when clicking to draft a Tweet, but it doesn't produce an event in the test dashboard on Countly and the event doesn't exist in the real dashboard. Do you recall what happened here?

Fixed: added `twitterShareTutorialPassed` custom event. logged when the user clicks on the twitter share link.

src/components/forms/feedback-survey/Form.vue
```diff
@@ -59,14 +59,11 @@
...
-        class="close"
+      <ButtonClose
         title="Dismiss survey"
-        v-on:click="onDismiss"
-        aria-label="Dismiss survey"
-      >
-        <img src="../../../static/images/close.svg" alt="dismiss or hide the tutorial feedback survey" />
-      </button>
+        :onDismiss="onDismiss"
+        imageAlt="Dismiss or hide the tutorial feedback survey"
```
terichadbourne, line 65:
```suggestion
        imageAlt="Dismiss tutorial feedback survey"
```

src/components/forms/feedback-survey/option-b/Form.vue
```diff
@@ -44,30 +44,27 @@
...
-        class="close"
+      <ButtonClose
         title="Dismiss survey"
-        v-on:click="onDismiss"
-        aria-label="Dismiss survey"
-      >
-        <img src="../../../../static/images/close.svg" alt="dismiss or hide the tutorial feedback survey" />
-      </button>
+        :onDismiss="onDismiss"
+        imageAlt="Dismiss or hide the tutorial feedback survey"
```
terichadbourne, line 66:
```suggestion
        imageAlt="Dismiss tutorial feedback survey"
```

src/components/forms/feedback-survey/Question.vue
```diff
@@ -1,39 +1,46 @@
...
+          >
+            <input
+              type="radio"
+              class="input-reset"
+              :value="index + 1"
+              :name="`response-${index + 1}`"
+              v-on:change="onSelect(index + 1)"
+              v-model="selected"
+            />
+            <img :src="icon" :alt="`answer ${index + 1} out of 5 to the question`" />
```
terichadbourne, line 33:
```suggestion
            <img :src="icon" :alt="`Rate as ${index + 1} on scale of 1 to 5`" />
```

terichadbourne (commented):
Looks great! Just added some suggestions for alt text edits. Feel free to incorporate and merge if you agree.

PR #572, opened 2020-11-26
fix: make sure the document, body and its children adapt to the viewport in height

zebateira:
ATM the content is not occupying the full viewport and the footer on the lesson pages does not stick to the bottom of the viewport on larger screens:

image

To fix this I adapted the layout styles from [next-with-moxy](https://github.com/moxystudio/next-with-moxy) to fix this issue thoroughly across browsers and devices.

Adapted from https://github.com/moxystudio/next-with-moxy/blob/master/www/shared/styles/global/layout.css

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -16,17 +16,48 @@ export default {
 
 <style>
 
-html {
-  overflow-y: scroll;
+*,
+*::before,
+*::after {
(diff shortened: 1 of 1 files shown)


terichadbourne (approved):
Looks great. Thank you!

PR #580, opened 2020-12-10
Add course pages

### LINKED ISSUE #579: Feature: Dedicated landing pages for courses
Breaking this off of #527 into its own issue. 

We need to create dedicated pages for courses, e.g. /courses/filecoin with dedicated text blocks/descriptions, which will be able to have dedicated social cards and other data to improve ranking in search engines. ([See here](https://github.com/ProtoSchool/protoschool.github.io/issues/526#issuecomment-674111396) how this is not possible with our current filter system on the tutorials page.) 

These will be used mostly as landing pages from web searches. They need a way for people to easily get out to the full tutorial listings but don't need the filtering system available in the full list. They'll have customized text blocks describing the content with words likely to end up in web searches, but the bottom portion of the page that pulls in tutorial cards will look identical to filtered tutorial listings. 

I've requested custom social cards for course pages in the design repo: https://github.com/protocol/design-shop/issues/441

I'll be taking the first stab at this using the dynamic implementation suggested by @zebateira in #527: 

> Notes on "Create dedicated pages for courses"
Content: create a new entry on static/translations/en.json for the necessary content
Creating the course pages: here we have two options: static or dynamic
2.1. Static: easiest and quickest. Basically create static pages for each course (IPFS, Multiformats, Filecoin, etc) and then add them to the static routes in the src/routes.js file.
2.2. Dynamic:
Use the courses.json file to generate the routes and feed it to a new entry in the routes file called courses which should be added to the all entry as well (all and featured need to be ignored from the list of courses from the courses.json file when generating the pages)
Create new pages/Course.vue page component that will render a generic course using the content of the translations file
Optional: re-structure the courses.json file to separate all and featured from the rest of the courses. something like ui: [all, featured], curated: [ipfs, filecoin, multiformats]

terichadbourne:
This WIP PR adds course landing pages for SEO purposes. See #579 for remaining action items. 

Partially addresses #579


src/pages/Course.vue
```diff
@@ -0,0 +1,164 @@
...
+      return this.courseFilter.tutorials.some(tutorialId => getTutorialType(tutorialId) === 'code' || getTutorialType(tutorialId) === 'file-upload')
+    },
+    courseDescription: function () {
+      return translations.courses[this.course].description
+    },
+    seoDescription: function () {
+      return translations.courses[this.course].seoDescription
+    },
+    otherCourses: function () {
+      return getAll().filter(course => course.id !== this.course && getCourseNames().includes(course.id))
```
zebateira, line 126: Here we are making the assumption that projects are somewhat equivalent to courses. Which is somewhat true in a way.

Although I think we could do better in making this distinction and make the actual data reflect the difference, but we can leave that for future work as a mini-refactor.

src/utils/projects.js
```diff
@@ -3,6 +3,7 @@ import data from '../static/projects.json'
 // Populate data with more properties
 data.forEach(project => {
   project.logo = require(`../static/images/projects/${project.id}.svg`)
+  project.path = `/course/${project.id}`
```
zebateira, line 6: yeah, I don't think this makes sense. going to revert this

zebateira (approved):
image

terichadbourne:
Merging MVP so we can start to collect SEO results. See #579 for improvements to make after the holidays.

PR #588, opened 2021-01-08
fix: move tutorial completed badge (trophy) to the top

zebateira:
Before:

image

After the fix:

image


terichadbourne:
Would love to turn this into a line drawing or use our checkmark icon once we update the cards to show the status as well, but totally agree about fixing this alignment. Thank you!

Issue #590, opened 2021-01-13
Build Error: EMFILE


zebateira:
Build is failing in Fleek (e.g. https://app.fleek.co/#/sites/protoschool/deploys/2021-01-13T07:40:48.705Z?accountId=protocollabs-team).

The testing server crashes with the error `EMFILE: Too many files open`, which indicates that the server process has opened too many files.
This implies that the server is doing too many OS Ops on files caused by too many requests from the tests.
There some optimisations we have pending, like changing the icons from url refs to inline (done herehttps://github.com/ProtoSchool/protoschool.github.io/pull/582), but only when restarting the testing server between browser test runs, does the build actually finish.

Several issues could be at play here, and these are the suspicions so far:

- Bug in the http server `serve`
- Bug in cypress browser cache config
- Content not being cached properly
- Very low `ulimit` for file descriptors

zebateira:
It seams the issue is the low `nofiles` value reported by `ulimits`.

PR #591, opened 2021-01-13
Fix/build fails emfile

zebateira:
Closes https://github.com/ProtoSchool/protoschool.github.io/issues/590


zebateira:
The fleek team increased the limit to

```
softLimit: 16384,
hardLimit: 131072,
```

This setting has been enabled for our project in fleek, so future docker images should also have this setting enabled.

PR #594, opened 2021-01-19
fix: positioning of radio button selected state

zebateira:
### Before

image

### After with fix

image


terichadbourne (approved):
Looks so traditionally correct. Sigh. 👀

Issue #597, opened 2021-01-22
Bug: Monaco Editor not using all features


zebateira:
Monaco editor is not working properly with all the necessary features, most importantly autocompletion and error highlighting.
We regressed on these features when we upgraded the monaco editor to a newer and more maintained vue component.

This is now causing UX issues for learners when using the coding tutorials (#596).

Issue #596, opened 2021-01-21
Lesson Feedback: Blog - Lesson 6 (List posts chronologically with a chain of links)


theNvN:
There might be a bug in the lesson 6 of  "Blogging on Decentralized Web" lesson.

URL: https://proto.school/blog/06

The Submitting of the solution to the exercise is not working. No errors/success message displayed....nothing happens.

Browser: Brave, Version 1.19.86

zebateira:
hello @theNvN!
Thank you for the bug report.

I've just tested this on Brave 1.19.86 and it worked with the solution code.
Do you mind sharing the code you submitted?

theNvN:
Hi @zebateira 
This is the code
```
/* globals ipfs */

const run = async () => {
  const natCid = await ipfs.dag.put({ author: "Nat" })
  const samCid = await ipfs.dag.put({ author: "Sam" })

  // Modify the blog posts below

  const treePostCid = await ipfs.dag.put({
    content: "trees",
    author: samCid,
    tags: ["outdoor", "hobby"]
  })
  const computerPostCid = await ipfs.dag.put({
    content: "computers",
    author: natCid,
    tags: ["hobby"]
    prev: treePostCid
  })
  const dogPostCid = await ipfs.dag.put({
    content: "dogs",
    author: samCid,
    tags: ["funny", "hobby"],
    prev: computerPostCid
  })

  const outdoorTagCid = await ipfs.dag.put({
    tag: "outdoor",
    posts: [treePostCid]
  })
  const hobbyTagCid = await ipfs.dag.put({
    tag: "hobby",
    posts: [treePostCid, computerPostCid, dogPostCid]
  })
  const funnyTagCid = await ipfs.dag.put({
    tag: "funny",
    posts: [dogPostCid]
  })

  return dogPostCid
}

return run
```

Code seems to process/load something (Submit btn is disabled for a moment), but no feedback, error or success, is outputted.

zebateira:
thank you for sharing that!
I can see the issue now: you're missing a comma in the `computerPostCid` variable creation, where `tags: ["hobby"]` should end with a comma. See final correct correct code:

```
/* globals ipfs */

const run = async () => {
  const natCid = await ipfs.dag.put({ author: "Nat" })
  const samCid = await ipfs.dag.put({ author: "Sam" })

  // Modify the blog posts below

  const treePostCid = await ipfs.dag.put({
    content: "trees",
    author: samCid,
[...]

PR #598, opened 2021-01-22
fix: show syntax error

zebateira:
Addresses #596 partially.

Syntax errors were not being given as feedback in the `Output` component on a specific case.
Although, the monaco editor should show the syntax error feedback right from the start to avoid situations like these #596 where the user simply missed a comma.

Fixing the code editor is filed here: https://github.com/ProtoSchool/protoschool.github.io/issues/597

This PR only addresses the issue of no error feedback when submitting code with a syntax error.

### DIFF
--- a/src/components/Lesson.vue
+++ b/src/components/Lesson.vue
@@ -176,8 +176,8 @@ const _eval = async (text, ipfs, args = []) => {
   try {
     // eslint-disable-next-line
     fn = new Function(Object.keys(modules).join(','), text)
-  } catch (err) {
-    return new SyntaxError(err.message, err)
+  } catch (error) {
+    return new SyntaxError(error.message, error)
   }
 
   let result
--- a/src/components/Output.vue
+++ b/src/components/Output.vue
@@ -1,15 +1,15 @@
 <template>
   <div class="pt2">
-    <div v-if="output.test.error"
-      class="lh-copy pv2 ph3 bg-red white"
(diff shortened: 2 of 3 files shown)


terichadbourne:
Thanks for this, @zebateira! 

Just tested this by making the same mistake the user did in issue #596 and got this output: 
image

Are we confident enough about the kinds of errors it will pick up (or do the errors come with extra properties we're not yet using) that we could make this say "Syntax Error: ..."? Does Monaco happen to provide line numbers we could pull in as well? It would be nice to have some way to flag that it's the code editor reporting this as opposed to us telling them they're messing up their usage of IPFS.

zebateira:
@terichadbourne pushed a fix so that the error message still shows the type of error before, eg, "SyntaxError: ..." instead of just "Error: ..."

> Does Monaco happen to provide line numbers we could pull in as well? It would be nice to have some way to flag that it's the code editor reporting this as opposed to us telling them they're messing up their usage of IPFS.

The line number property is not a standard: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/lineNumber
In any case, I wouldn't bother with this because we need to fix the monaco editor integration to show this information, because it already supports it: https://microsoft.github.io/monaco-editor/

terichadbourne (approved):
Looks great. Thanks for making that change.

PR #612, opened 2021-02-15
fix macos build: update ipfs dependency to v0.50.2

### LINKED ISSUE #576: Bug: IPFS not installing when repo is cloned and installed
In testing, multiple non-coder users have been unable install our dependencies after cloning the repo for the first time, with errors showing that it's IPFS that isn't installing properly. Both had Macs and had installed NPM from its website. 

@zebateira is working on debugging.

yusefnapora:
I wasn't able to build the project when first checking this repo out. I'm running on an M1 MacBook Air running Big Sur (version 11.1). 

I think this error is related to old versions of some dependencies being pinned in `package-lock.json` and forcing an old version of `node-gyp`, which fails on recent builds of Mac OS.

I was able to regenerate the `package-lock.json` without updating any dependencies in `package.json`, and now the build succeeds. The `bcrypto` package that was failing to build on the main branch isn't even pulled in anymore, so one of our transitive dependencies must have updated.

Hopefully closes #576

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4941,15 +4941,6 @@
         "tweetnacl": "^0.14.3"
       }
     },
-    "bcrypto": {
-      "version": "5.1.0",
-      "resolved": "https://registry.npmjs.org/bcrypto/-/bcrypto-5.1.0.tgz",
-      "integrity": "sha512-WEs5g7WHdEdLLcsvhE7Z1AXv0G+hb+bJhSUYM7samFNrH051XhcFVWxAbAZDmIU1HWjpjhmQ+HqBar7UC/qrzQ==",
-      "requires": {
-        "bufio": "~1.0.6",
-        "loady": "~0.0.1"
-      }
-    },
     "bech32": {
       "version": "1.1.4",
       "resolved": "https://registry.npmjs.org/bech32/-/bech32-1.1.4.tgz",
@@ -5606,11 +5597,6 @@
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/buffer-xor/-/buffer-xor-1.0.3.tgz",
       "integrity": "sha1-JuYe0UIvtw3ULm42cp7VHYVf6Nk="
-    },
-    "bufio": {
-      "version": "1.0.7",
-      "resolved": "https://registry.npmjs.org/bufio/-/bufio-1.0.7.tgz",
-      "integrity": "sha512-bd1dDQhiC+bEbEfg56IdBv7faWa6OipMs/AFFFvtFnB3wAYjlwQpQRZ0pm6ZkgtfL0pILRXhKxOiQj6UzoMR7A=="
     },
     "builtin-status-codes": {
       "version": "3.0.0",
@@ -16595,7 +16581,6 @@
       "resolved": "https://registry.npmjs.org/libp2p-noise/-/libp2p-noise-1.1.2.tgz",
       "integrity": "sha512-iKXdzGnPsz3slh6Gm9oNj0h0X37f/YFuSkg7MikQgrx5l5XRaFRxVDoqbsTlQ5nIS02tGuLJvmbqpLOZ+aWVow==",
       "requires": {
-        "bcrypto": "5.1.0",
         "buffer": "^5.4.3",
         "debug": "^4.1.1",
         "it-buffer": "^0.1.1",
@@ -17266,11 +17251,6 @@
           }
         }
       }
-    },
-    "loady": {
-      "version": "0.0.5",
-      "resolved": "https://registry.npmjs.org/loady/-/loady-0.0.5.tgz",
-      "integrity": "sha512-uxKD2HIj042/HBx77NBcmEPsD+hxCgAtjEWlYNScuUjIsh/62Uyu39GOR68TBR68v+jqDL9zfftCWoUo4y03sQ=="
     },
     "local-web-server": {
       "version": "4.2.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1292,109 +1292,6 @@
         }
       }
     },
-    "@grpc/grpc-js": {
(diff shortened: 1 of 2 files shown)


yusefnapora:
BTW, I also hit this issue related to the new M1 macs: https://github.com/ProtoSchool/protoschool.github.io/issues/613

But it's a pretty edgy edge case and probably won't affect that many people.

terichadbourne:
Thank you @yusefnapora! I'm going to defer to @zebateira on reviewing this one since he's been more involved in investigating our earlier issues with `npm install`.

zebateira:
The build stills fails 😢 

image

Which is weird because I also see this happening:

> The bcrypto package that was failing to build on the main branch isn't even pulled in anymore, so one of our transitive dependencies must have updated.

image

Looking into this 👍

yusefnapora:
Hmm, I think I may be making things more confusing by having two versions of nodejs installed. If I completely delete `node_modules`, I get different results for `npm ls bcrypto` depending on which node I use.

With version 15.8.0, compiled for apple silicon:

```
$ env PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 npm install

added 3008 packages, and audited 3075 packages in 22s

33 vulnerabilities (3 low, 4 moderate, 26 high)

To address issues that do not require attention, run:
  npm audit fix

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.
```

```
$ npm ls bcrypto
protoschool@1.0.0 /Users/yusef/work/repos/protoschool.github.io
└── (empty)

```


But, if I install with v14.15.5 (Intel version), I get this:

```
npm ls bcrypto
protoschool@1.0.0 /Users/yusef/work/repos/protoschool.github.io
└─┬ ipfs@0.48.2
  └─┬ libp2p-noise@1.1.2
    └── UNMET DEPENDENCY bcrypto@5.1.0

npm ERR! missing: bcrypto@5.1.0, required by libp2p-noise@1.1.2
```

as well as a warning about the package lock file using version 2 of the lockfile syntax when I do the `npm install`:

```
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
```

So it looks like the "fix" from this branch only works with node v15 and the new lockfile syntax, although the install still succeeds using the v2 lockfile (on my machine).

I just added a commit that bumps our `ipfs` dependency to the latest, v0.54.2 - this should solve the root problem, since now we depend on a non-broken version of bcrypto (5.4.0): 

```
npm ls bcrypto
protoschool@1.0.0 /Users/yusef/work/repos/protoschool.github.io
└─┬ ipfs@0.54.2
  └─┬ ipfs-core@0.5.2
    └─┬ libp2p-noise@2.0.4
      └── bcrypto@5.4.0
```

Assuming that updating ipfs doesn't break anything, we should be good :)

yusefnapora:
btw, I forgot to mention that I used v14 for the previous commit, so it has the version 1 lockfile.

yusefnapora:
> Assuming that updating ipfs doesn't break anything, we should be good :)

hmm... I'll have a go at running the tests locally and see if anything jumps out at me

zebateira:
> Assuming that updating ipfs doesn't break anything, we should be good :)

Unfortunately updating IPFS in ProtoSchool is not as straightforward as it could be: because our code challenges and content directly depend on a specific version of IPFS, we need to carefully analyze each IPFS release and check the impact on the code and content. Code migrations are easier than content updates.

But this is a good call out for this issue since there are somethings that we can improve to make this process more streamlined:

 - Document IPFS usage dependency on content and code
 - Setup a scheduled IPFS update task so we don't fall behind so often and risk going through issues like these 

Tracking this [here](https://github.com/ProtoSchool/protoschool.github.io/issues/616). It also takes into account that libp2p might need this as well.

zebateira:
@yusefnapora can you make your local build work without updating IPFS? We can temporarily only use node v15, work on the new libp2p tute, and then we can come back to this and upgrade IPFS.
If not, upgrading IPFS is the solution anyways because we will need to do it sometime soon.

zebateira:
I can't seem to make it work in node v15. And since it doesn't work in fleek anyways, I guess we need to tackle the IPFS dep update.

yusefnapora:
I think I found a version if IPFS that's new enough to fix the issue, but not new _enough_ to cause it's own issues 😄 

That is, `ipfs@0.50.2` pulls in `bcrypto@5.4.0`, which builds fine on my machine. I'm still running the tests locally, but the file upload tests that were failing with the latest `ipfs` have passed, so I've pushed it up for fleek to try. Fingers crossed...

yusefnapora:
Yay, the tests passed locally!

```
  (Run Finished)


       Spec                                              Tests  Passing  Failing  Pending  Skipped
  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ ✔  courses.spec.js                          00:06        4        4        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✔  redirects.spec.js                        00:18        1        1        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✔  tutorial-messages.spec.js                00:12       14       14        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✔  tutorial-redirect-modal.spec.js          00:23       12       12        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✔  tutorials.spec.js                        11:44      338      338        -        -        - │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
    ✔  All specs passed!                        12:44      369      369        -        -        -
```

zebateira:
@yusefnapora yay!

zebateira (approved):
💯

PR #620, opened 2021-03-01
Add warning for mobile users attempting coding tutorials

### LINKED ISSUE #147: Lesson Feedback: Basics - Lesson 1 (Create a node and return a Content Identifier (CID))
Have a question or suggestion regarding a specific ProtoSchool lesson? Please use this template to share it!

**URL of the lesson that's confusing:**
eg https://proto.school/#/blog/01

**What's confusing about this lesson?**
The code editor isn't a text box, it's something that is broken on android. 

**What additional context could we provide to help you succeed?**
Have a next button that you can use when the code editor is buggy

**What other feedback would you like to share about ProtoSchool?**

terichadbourne:
Closes #147 by warning users with low res devices that they won't be able to complete coding challenges without a keyboard.

### DIFF
--- a/src/components/Lesson.vue
+++ b/src/components/Lesson.vue
@@ -2,6 +2,7 @@
   <div :class="{'overflow-hidden': expandChallenge}">
     <TutorialRedirectModal :tutorial="tutorial" :lesson="lesson" />
     <MobileWarningModal v-if="!isResources && (tutorialType === 'code' || tutorialType === 'file-upload')"
+      targetClass="dn-ns"
       :tutorial="tutorial"
       :lesson="lesson"
     />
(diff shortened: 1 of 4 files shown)


terichadbourne:
Working on a banner to use on the landing page of coding tutorials and on the non-coding lessons that are sometimes included in coding tutorials. How do you feel about this concept, @zebateira? If we included messaging about both the JavaScript and the device, we could use it across the board on these to also help warn folks who don't know how to code. We could also cut up the text conditionally so that it only mentions the device if our media query suggests your device won't work. 

on landing page of coding tutorial

image

on non-coding lesson w/in coding tutorial
image

zebateira:
Looking good 👍 

Some concerns:

- Will it look too cramped when the a tutorial message shows up for new / updated tutorials? Would like to see that.
- Maybe it makes more sense to move the warning above the lesson title in the second case.

PR #632, opened 2021-03-22
Update emojis in Merkle DAG images to be distinct

### LINKED ISSUE #592: Update emoji in Merkle DAG tutorial diagrams
Fix diagrams in Merkle DAGs tutorial so emoji aren't reused. For more see comments in #582: https://github.com/ProtoSchool/protoschool.github.io/pull/582#issuecomment-759860966

@mitchwagner I believe you're on top of this, but let me know if you need someone else to step in. Thanks!

mitchwagner:
Resolves #592

I additionally re-did the original images to be SVGs instead of PNGs, so that they scale better.

Note: I still need to update the text to reflect these changes!


terichadbourne:
@mitchwagner Thanks so much for this! Just let me know when you've made the necessary text edits and I'll take closer look. Much appreciated!

terichadbourne:
@mitchwagner Just confirming you still have work left to do on this one, right?

mitchwagner:
@terichadbourne Yep, my apologies - I haven't gotten around to the text edits. I will rebase and update the PR this weekend.

mitchwagner:
I've adjusted the text. Some of the images in Lesson 4 could probably stand to be made a bit smaller - do we have any easy way of achieving that?

terichadbourne:
Thanks for this @mitchwagner! I was able to use HTML instead of Markdown on a couple of the images to set a max width. (Screenshot below.) 

However, we noticed that the emojis should actually be inside the closing curly brackets, not outside. Is that something you could update for us?

image

mitchwagner:
> However, we noticed that the emojis should actually be inside the closing curly brackets, not outside. Is that something you could update for us?
> 
Allow me to investigate - I think I missed something with the SVGs. I may need to ensure I'm embedding the assets (including font) properly, because they definitely should not look like that (below is what I see in a browser).

Screen Shot 2021-04-29 at 11 32 30 PM

terichadbourne:
The new images look great, @mitchwagner, thank you! Just doubling checking that you've taken care of everything you planned to before I merge?

terichadbourne:
Thanks for confirming, @mitchwagner! @zebateira is going to do some image compression magic and then we'll get this merged!

terichadbourne:
Yes, looks good. Thank you both for all your work on this!

PR #653, opened 2021-04-30
Automatically trigger weekly deployments on fleek

zebateira:
This will make sure that events stay up to date and mailchimp subscriptions are triggered.
@terichadbourne let me know if weekly cadence is enough. It's set to trigger every Monday at 9am UTC.


zebateira:
update to trigger everyday at 9am UTC

.github/workflows/weekly-build.yml
```diff
@@ -0,0 +1,17 @@
+name: Weekly Build (events and mailchimp subscriptions updates)
```
terichadbourne, line 1:
```suggestion
name: Daily Build (events and Mailchimp subscription updates)
```

terichadbourne (changes requested):
Would suggest updating the name to match the new schedule we discussed today. Anything you need me to do to test this one? If not, please go ahead and merge when ready!

.github/workflows/weekly-build.yml
zebateira: this is it! will merge

PR #720, opened 2021-09-02
chore: update docs related to remove events data

zebateira:


scripts/README.md
```diff
@@ -67,7 +67,13 @@ The production values for the variables are set up in Fleek's build UI.
 
 #### `npm run scripts:build:data`
 
-As part of the build process (see [fleek config](../.fleek.json#L11)), we fetch all event data for submitted events and write the data for approved events to specific `static/*.json` files so the UI can read these JSON files and render the content as event listings. Additionally, we add new event organizers to our Mailchimp audience to subscribe them to our newsletter.
+This script is run as part of the build process (see [fleek config](../.fleek.json#L11)).
+It should be used to affect any data of the website that you might want to change automatically based on remote sources.
+
+Currently, only the events list is based on remote data (google spreadsheet events file), and the script performs the following actions:
```
terichadbourne, line 73:
```suggestion
Currently, only the events list is based on remote data (a Google spreadsheet generated from our workshop submission form), and the script performs the following actions:
```

scripts/README.md
```diff
@@ -67,7 +67,13 @@ The production values for the variables are set up in Fleek's build UI.
...
+
+- fetch all event data for submitted events from the google spreadsheet events list and write the data for approved events to specific `static/*.json` files so the UI can read these JSON files and render the content as event listings.
```
terichadbourne, line 75:
```suggestion
- fetches all event data for submitted events from the Google Sheets and writes the data for approved events to specific `static/*.json` files so the UI can read these JSON files and render the content as event listings. (All entries on the sheet must be manually approved before the script will fetch them.) 
```

scripts/README.md
```diff
@@ -67,7 +67,13 @@ The production values for the variables are set up in Fleek's build UI.
...
+- use the email addresses of the users that submitted the events and add them as new event organizers to our Mailchimp audience to subscribe them to our newsletter. This is only done if they aren't subscribed yet and only if they have checked the box to accept the subscription when submitting the event.
```
terichadbourne, line 76:
```suggestion
- takes the email addresses of the users that submitted the events and adds them to our Mailchimp audience, subscribing them to our local leadership newsletter. This is only done if they aren't subscribed yet, and after they've acknowledged through the form submission that they agree to receive these updates.
```

terichadbourne (changes requested):
Thanks so much for these updates! Added a few suggestions; feel free to merge if you're game to accept them.

scripts/README.md
zebateira: The script will fetch all the events, and then only use approved ones. So maybe we could say `before the script will use them`?

zebateira:
```suggestion
- fetches all event data for submitted events from the Google Sheets and writes the data for approved events to specific `static/*.json` files so the UI can read these JSON files and render the content as event listings. (Only manually approved event entries will be used by the script and added to the JSON files.) 
```

PR #791, opened 2022-04-13
updated broken link about sector sizes

walkerlj0:
filecoin docs have been shuffled a lot, linked to filecoin spec instead


yiannisbot (approved):
Great catch! Thanks!

PR #880, opened 2023-08-23
feat(es6): :arrow_up: Upgrade to ESM

whizzzkid:


ElPaisano:
@whizzzkid thanks for creating this

I pulled down and tested locally. Wasn't ablt to get it to build

